PR #19, opened 2017-07-06
Compatibility fixes for GCC 4.8.2

ariddell:
Addresses #19


ariddell:
"Every expression is either an lvalue or an rvalue" madness!

PR #22, opened 2017-07-07
Use sqlite3 instead of lmdb

ariddell:
sqlite3 comes with Python. lmdb does not. The difference in performance
is almost surely negligible for all uses in httpstan.


ariddell:
Merged in manually.

PR #23, opened 2017-07-08
Add -stdlib=libc++ for clang

ariddell:


ariddell:
Perhaps not needed? At least for a certain version of OS X and clang.

PR #24, opened 2017-07-11
sqlite3 fixes: turn on write-ahead-log

ariddell:
By default sqlite3 optimizes for compatibility. We can assume that a
user with Python 3.6 available has a non-ancient version of sqlite3.


ariddell:
Failing test has to do with mypy.

PR #44, opened 2018-06-06
Draw samples in parallel using system threads.

riddell-stan:
Relatively minor change. Sets STAN_THREADS macro and removes lock.

### DIFF
--- a/httpstan/cache.py
+++ b/httpstan/cache.py
@@ -32,6 +32,7 @@ async def init_cache(app):
     cache_path = appdirs.user_cache_dir("httpstan")
     os.makedirs(cache_path, exist_ok=True)
     logging.info(f"Opening cache in `{cache_path}`.")
+    # if `check_same_thread` is False, use of `conn` across threads should work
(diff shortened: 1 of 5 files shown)


ahartikainen:
This will probably not fix the CI error but it should remove the warning.

In `tox.ini` edit the line

    [testenv:format]
    deps =
        black==18.6b1
    commands = bash -c "find httpstan tests scripts -not -path 'httpstan/lib/*' -name '*.py' -type f -print0 | xargs -0 black --py36 --check --line-length=99"    

to

    [testenv:format]
    deps =
        black==18.6b1
    whitelist_externals = bash
    commands = bash -c "find httpstan tests scripts -not -path 'httpstan/lib/*' -name '*.py' -type f -print0 | xargs -0 black --py36 --check --line-length=99"

ahartikainen:
Yeah, I did more reading about the black documentation.

So the error was only due to `--check` with two files that need changing. So probably nothing to worry about.

riddell-stan:
Thanks. This is all still a bit rough. The master branch has the formatting errors. I think.

PR #59, opened 2018-06-26
Appveyor setup - msvc + clang-cl

ahartikainen:
clang-cl + appveyor

old - mingw-w64 toolchain + appveyor miniconda setup

Runs tests with python subprocess, ignores crash at exit.

### DIFF
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -8,11 +8,11 @@ import functools
 import hashlib
 import importlib
 import os
-import shutil as _shutil
+import shutil
 import string
 import sys
 import tempfile
-import warnings as _warnings
+import warnings
 from typing import List
 from typing import Optional
 from typing import Tuple  # noqa: flake8 bug, #118
@@ -35,12 +35,12 @@ class TemporaryDirectory(tempfile.TemporaryDirectory):
 
     @classmethod
     def _cleanup(self, name, warn_message):
-        _shutil.rmtree(name, ignore_errors=True)
-        _warnings.warn(warn_message, ResourceWarning)
+        shutil.rmtree(name, ignore_errors=True)
+        warnings.warn(warn_message, ResourceWarning)
 
     def cleanup(self):
         if self._finalizer.detach():
-            _shutil.rmtree(self.name, ignore_errors=True)
+            shutil.rmtree(self.name, ignore_errors=True)
 
 
 def calculate_model_id(program_code: str) -> str:
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -8,9 +8,12 @@ import functools
 import hashlib
 import importlib
 import os
+import shutil as _shutil
+import stat
 import string
 import sys
 import tempfile
+import warnings as _warnings
 from typing import List
 from typing import Optional
 from typing import Tuple  # noqa: flake8 bug, #118
@@ -23,6 +26,32 @@ import pkg_resources
 
 import httpstan.compile
 import httpstan.stan
+
+
+# TemporaryDirectory problem with Windows
+# see https://docs.python.org/3/library/shutil.html?highlight=shutil#rmtree-example
+class TemporaryDirectory(tempfile.TemporaryDirectory):
+
+    def __init__(self, suffix=None, prefix=None, dir=None):
+        super(TemporaryDirectory, self).__init__(suffix, prefix, dir)
+
+    def _remove_readonly(self, func, path, _):
+        "Clear the readonly bit and reattempt the removal"
+        os.chmod(path, stat.S_IWUSR|stat.S_IREAD)
+        try:
+            func(path)
+        except (PermissionError, OSError):
+            # WIP
+            pass
+
+    @classmethod
+    def _cleanup(self, name, warn_message):
+        _shutil.rmtree(name, onerror=self._remove_readonly)
+        _warnings.warn(warn_message, ResourceWarning)
+
+    def cleanup(self):
+        if self._finalizer.detach():
+            _shutil.rmtree(self.name, onerror=self._remove_readonly)
 
 
 def calculate_model_id(program_code: str) -> str:
@@ -151,14 +180,17 @@ def load_model_extension_module(model_id: str, module_bytes: bytes):
     # of this function will not load.
     module_name = calculate_module_name(model_id)
     if sys.platform.startswith("win"):
-        module_filename = f"{module_name}.pyd"
+        module_filename = f"{module_name}.cp36-win_amd64.pyd"
     else:
         module_filename = f"{module_name}.so"
     with tempfile.TemporaryDirectory() as temporary_directory:
         with open(os.path.join(temporary_directory, module_filename), "wb") as fh:
             fh.write(module_bytes)
         module_path = temporary_directory
-        assert module_name == os.path.splitext(module_filename)[0]
+        if sys.platform.startswith("win"):
+            assert module_name + '.cp36-win_amd64' == os.path.splitext(module_filename)[0]
+        else:
+            assert module_name == os.path.splitext(module_filename)[0]
         return _load_module(module_name, module_path)
 
 
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -283,6 +283,6 @@ def _build_extension_module(
                 return fh.read()
         else:
             module = _load_module(module_name, build_extension.build_lib)
-            with open(module.__file__, "rb") as fh:
-                assert module.__name__ == module_name, (module.__name__, module_name)
+            assert module.__name__ == module_name, (module.__name__, module_name)
+            with open(module.__file__, "rb") as fh: 
                 return fh.read()  # type: ignore  # pending fix, see mypy#3062
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -31,27 +31,17 @@ import httpstan.stan
 # TemporaryDirectory problem with Windows
 # see https://docs.python.org/3/library/shutil.html?highlight=shutil#rmtree-example
 class TemporaryDirectory(tempfile.TemporaryDirectory):
-
     def __init__(self, suffix=None, prefix=None, dir=None):
         super(TemporaryDirectory, self).__init__(suffix, prefix, dir)
 
-    def _remove_readonly(self, func, path, _):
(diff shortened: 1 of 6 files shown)


ahartikainen:
@ariddell do you need to somehow enable appveyor? It works on my appveyor setup.

riddell-stan:
I had disabled appveyor, now it's back.

This looks great. Thanks! Would you mind adding a note or two in the commit or in the docstrings explaining what's going on with ``D_hypot`` and with the need for _find_module under windows?

ahartikainen:
I added some explanations for the source.

`-D_hypot=hypot` fixes a bug in mingw-w64 compiler (typo in the source). It should not affect *nix systems anyway (links function to "empty" variable which is not then used).

There still are some problem with tests (or compilation), I have to sort that out.

ahartikainen:
Good news are:

    building + sampling works
    threaded sampling works

WIP:

    Can not delete temporary files if loaded with importlib.
    Some problem with the "dummy" module 
        test_arguments: after fixing the suffix there is some error:
            ImportError: DLL load failed: %1 is not a valid Win32 application. 
    module needs specific suffix to be importable. ".cp36-win_amd64"
    pytest or something is crashing the python on exit.

riddell-stan:
Sounds like progress. Thanks for your work on this.

ahartikainen:
This is still the problem: `pytest or something is crashing the python on exit.`

I don't know if this is related to problem that one can not remove temporary file if it is imported. 
Somehow some thread is failing at exit and then everything else crashes (something is trying to access memory released).

Not even sure if this problem is only due to py.test?

Do we have some example of running httpstan with normal python. I did have some example long time a go. Or should I try these with `pystan-next`?

riddell-stan:
I went hunting for the exit code here:

```
[00:24:12] Command exited with code -1073740940
```

and found these:

- https://github.com/scikit-image/scikit-image/issues/3090
- https://github.com/statsmodels/statsmodels/pull/3834

It's great that the tests are passing. Looks like we should at least be able to build windows wheels.

httpstan/models.py
```diff
@@ -109,6 +127,23 @@ def _load_module(module_name: str, module_path: str):
     return module
 
 
+def _find_module(module_name: str, module_path: str):
```
riddell-stan: Could we use this, `_find_module` instead of `__load_module`? It would be nice to only maintain one of these.

httpstan/models.py
```diff
@@ -229,7 +275,14 @@ def _build_extension_module(
...
-        module = _load_module(module_name, build_extension.build_lib)
-        with open(module.__file__, "rb") as fh:  # type: ignore  # pending fix, see mypy#3062
+        if sys.platform.startswith("win"):
+            # tempfile bug: Windows can't open file that is already open
+            # see related statement in https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile
+            path = _find_module(module_name, build_extension.build_lib)
+            with open(path, "rb") as fh:  # type: ignore  # pending fix, see mypy#3062
+                return fh.read()  # type: ignore  # pending fix, see mypy#3062
+        else:
+            module = _load_module(module_name, build_extension.build_lib)
```
riddell-stan: see comment above. I think _find_module (or maybe renamed, _find_module_path) might replace the old _load_module

httpstan/models.py
```diff
@@ -8,9 +8,11 @@
 import hashlib
 import importlib
 import os
+import shutil as _shutil
```
riddell-stan: Would you mind telling me why these are imported with underscore-prefixed aliases?

riddell-stan:
The travis error doesn't look serious, #66 should fix that.

httpstan/models.py
ahartikainen: Just a copy-paste error

httpstan/models.py
ahartikainen: Yes, we could

ahartikainen:
There is still something that get stuck randomly on appveyor. 

On my machine tests are working.

riddell-stan:
I think it's hitting the 1 hour limit on appveyor. Can we run a subset of the tests?

httpstan/models.py
ahartikainen: Is it ok to remove the assert (it tests againts loaded module)

     assert module.__name__ == module_name, (module.__name__, module_name)

ahartikainen:
I renamed `_find_module` to `_find_module_path` .

`_load_module` is still needed in `load_model_extension_module` function so I can not remove it.

riddell-stan:
I prefer `_find_module_path`. Good choice!

This all looks good. And it works! Thanks for all the hard work.

I am a little frustrated by all the hacks needed to get things working on Windows. If you have any ideas for simplifying things, feel free to propose them. For example, I wonder if we could make the code much simpler by simply not deleting things on Windows. (So Windows users would have to clear their temporary directories themselves.)

Let's merge this. Would you mind rebasing and squashing the ~30 commits into one or three commits?

ahartikainen:
Ok, currently there are no real solution for the python crash at exit. We just either need to make a (helper?) function to disable the crash reporting at exit or give a warning for users.

https://github.com/Alexpux/MINGW-packages/issues/2519

https://sourceforge.net/p/mingw-w64/bugs/445/

riddell-stan:
Does this mean that pystan will always crash for Windows users (and show a crash dialog window)?

If so we could perhaps disable multithreading for Windows users.

Or maybe we should try looking at the clang Windows compiler? Has anyone in the conda world tried experimenting with that?

ahartikainen:
Clang-cl could work. It is very new, but clang-cl  should be available in appveyor.

ahartikainen:
I can confirm that `httpstan` can be compiled with `clang-cl` and `httpstan` is able to compile working models with `clang-cl`. All tests pass. Not sure if `thread_local` works, but at least it compiles and samples correctly.

I installed clang with conda

    conda install llvmdev -c conda-forge

The problem is that `clang-cl` is not yet supported by `distutils`/`setuptools`.

I changed `*.h` files shown in the CPython PR https://github.com/python/cpython/pull/7680

Also I edited the `distutils\_msvccompiler.py` the line 245 

    self.cc = _find_exe("cl", paths)

to 

    self.cc = _find_exe("clang-cl.exe", paths) 

Also `distutils.cfg` should have (if exists)

    [build]
    compiler=msvc


Should we wait for `distutils` / `setuptools` to update their files?

riddell-stan:
This is good news! We probably should wait until distutils/setuptools supports it. With luck the patch will be merged before too long.

ahartikainen:
The current PR in CPython does not touch distutils. So they will need to make another PR for that.

`clang-cl` needs a MSVC C++ toolkit from the microsoft website, so it needs admin rights for installation. 

I'm not sure if `clang` interface needs external libraries.

Also I hope they would define `clang-cl` and `clang` options for `distutils.ccompiler`, so one could either use MSVC interface of GCC interface.  

I have locally created a new class for `clang-cl` under `distutils._msvc*` (and edited `distutils.ccompiler`), but it still quite hacky. That way `distutils.cfg` can use `compiler=clang-cl`.

I think we could create an option for user to use `mingw` and output a warning that atexit this will fail if one uses `parallel (thread_local)` execution and force remove the error window. Basically the error does not influence anything inside Stan executation, only when Python exists, it will fail. This way we have a way to run httpstan on Windows before `clang-cl` gets implemented in distutils/setuptools.

I hope `mingw / winpthreads` folks could fix the problem (at least the loop order) so `winpthreads` would not fail.

riddell-stan:
Should we perhaps just disable parallel sampling in Windows for now? That way we could still use pytest and build Windows wheels without issue. People could still use PyStan on Windows for tutorials and education but wouldn't be (really) able to use it for research.

ahartikainen:
Parallel sampling could be optional (I'm thinking environmental variable) on Windows.

    parallel = bool(int(sys.environ.get("HTTPSTAN_PARALLEL", "0")))

That way advanced users can easily use it

    import
    sys.environ["HTTPSTAN_PARALLEL"] = "1"

riddell-stan:
That does sound good. How about HTTPSTAN_WINDOWS_PARALLEL (since others 
will be parallel by default).

riddella:
@ahartikainen I'm thinking of taking another stab at Windows support. If we do not compile with STAN_THREADS on Windows we can use mingw, right?

riddell-stan:
@ahartikainen thinking of a different approach for Windows. If we disable STAN_THREADS on Windows we can use mingw without problems, right?

ahartikainen:
I believe it wont take too long before clang-cl is supported.

Mingw-w64 works. It just crashes when one exists the whole session. So everything is still saved and works as far as Stan is concerned. That could be fixed, it is doable.

This is also probably affecting RStan users on Windows?

Then there is still clang option on Windows, but I believe it also needs update for distutils.

riddell-stan:
This sounds good. I'll work on making a minimum viable version of pystan-next / httpstan without any support for Windows. Once clang works we can add it in.

riddell-stan:
@ahartikainen Thanks for your work on this. I agree it would be great to test on Windows.

I am annoyed that we have to change ~ 27% of the lines of code in ``models.py`` to support Windows. Having all this specialized code is not just annoying to non-Windows users but I think it reduces the readability of the code in general -- this, in turn, probably hurts maintainability.

Can we break up the work into smaller PRs? For example,

- PR #1 enables Windows model compilation (but nothing else) and runs the compilation-specific tests on appveyor. This wouldn't need to use clang-cl or the thread crashing dialog protection, right? So the PR would be simpler. (And maybe we could figure out a way to cleanup the TemporaryDirectory stuff.)
- PR #2 uses clang-cl and runs the full test suite.

What do you think of this plan?

(Of course, I'm also frustrated that there isn't an existing, widely-supported windows compiler that supports C++11 threading! The year, last time I checked, is 2018.)

ahartikainen:
I have not followed on this, but if I doubt anything has changed.

It is somehow a template problem with msvc. Clang-cl can fix it and I hope distutils could use it. Numba uses clang, so maybe they have done something differently.

Official Eigen Windows guides show totally different format for templates, so I don't know how much needs fixing.

ahartikainen:
I will close this now in favor of #142

PR #76, opened 2018-10-16
use uvloop instead of cpython event loop

riddell-stan:
May speed things up.


riddell-stan:
Appears to speed things up a bit.

PR #111, opened 2018-11-19
doc: add coverage badge

riddell-stan:


riddell-stan:
If/when the badge works, this can be merged. It doesn't work right now. Might be waiting on coveralls.io.

PR #144, opened 2019-03-03
[WIP] test: add windows CI config, appveyor CI

riddell-stan:


riddell-stan:
Since appveyor images now include miniconda, I think we can simplify things relative to the old pystan appveyor setup.

PR #147, opened 2019-03-03
fix: MinGW specific fix, cmath expects hypot

riddell-stan:


riddell-stan:
An explanation of this particular issue may be found at https://stackoverflow.com/questions/10660524/error-building-boost-1-49-0-with-gcc-4-7-0/12124708#12124708 

There's some sort of macro in Python(.h?) which creates a problem with MinGW.

PR #151, opened 2019-03-04
[WIP] Windows support (MinGW)

riddell-stan:
This updates #142 to the current `master`.

`STAN_THREADS` is defined on Windows. My hope is that non-parallel sampling can be made to work without crashing.

This PR is still a WIP. There's some minor cleanup to do if things work.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -37,5 +37,4 @@ build: false
 test_script:
   - mkdir empty_folder
   - cd empty_folder
-  # eventually `- python -m pytest ../tests`
-  - python -c "import httpstan"
+  - python -m pytest ../tests/test_bernoulli.py
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -330,4 +330,4 @@ def _build_extension_module(
         assert module.__name__ == module_name, (module.__name__, module_name)
         with open(module.__file__, "rb") as fh:  # type: ignore  # see mypy#3062
             module_bytes = fh.read()  # type: ignore  # see mypy#3062
-    return module_bytes, compiler_output
+        return module_bytes, compiler_output
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -12,6 +12,7 @@ import logging
 import os
 import pathlib
 import platform
+import shutil
 import string
 import sys
 import tempfile
@@ -29,6 +30,18 @@ import httpstan.compile
 import httpstan.stan
 
 logger = logging.getLogger("httpstan")
+
+
+class TemporaryDirectory(tempfile.TemporaryDirectory):
+    """Patch TemporaryDirectory to ignore errors."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    def cleanup(self):
+        """Ignore errors"""
+        if self._finalizer.detach():
+            shutil.rmtree(self.name, ignore_errors=True)
 
 
 def calculate_model_name(program_code: str) -> str:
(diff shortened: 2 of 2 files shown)


riddell-stan:
Nice idea for cleanup: "I also think we could add a script that tries to delete old httpstan tempfiles (httpstan_temp_ prefix) at start."

At this point, I'm more than happy to fill up the hard drive if it means we can support Windows. We can deal with cleanup in a future PR.

riddell-stan:
This looks good. Sampling works fine, except for that infamous `Command exited with code -1073740940`.

ahartikainen:
A few weeks ago I had a question/conversation with Ray Donnelly (@mingwandroid) about this problem in conda-forge gitter:

```
Ari Hartikainen @ahartikainen Feb 04 01:21
Hi, does anyone know if M2W64-toolchain will be updated to 6.0.0? Or even where to ask this?

Ray Donnelly @mingwandroid Feb 04 10:20
@ahartikainen we have no real plans for that at present (it could be done, but there may be better options) . What from GCC 6 do you need?

Ari Hartikainen @ahartikainen Feb 04 14:12
@mingwandroid I'm looking for a working thread-local implementation on Windows. clang-cl + msvc has one, but msvc needs admin and a patch for distutils.
This is to enable threading stuff with Stan (PyStan)
(RStan goes with RTools 4.0 which has gcc/g++ 8)

Ray Donnelly @mingwandroid Feb 04 15:40
@ahartikainen thread local works with GCC 5 on Windows. Do you have a test-case showing otherwise?

Ari Hartikainen @ahartikainen Feb 04 19:11
@mingwandroid hmm, it does work but segfaults at exit. I can try to create an example for you.

Ari Hartikainen @ahartikainen Feb 04 19:11
(with m2w64-toolchain)

Ray Donnelly @mingwandroid Feb 04 19:12
@ahartikainen are you passing -static or -static-libgcc?

Ari Hartikainen @ahartikainen Feb 04 19:51
@mingwandroid tbh I don't know. We call distutils, so I need to check logs
```

I have been a bit busy lately, but do we have some bare bones version of this problem?

cc. @wds15

riddell-stan:
I'm tempted to see what the error code is (if any) if we use the new 
thread-local storage PR. I'll work on that.

wds15:
I am really curious if the new PR for threading works for you just fine! However, be prepared to have to do (hopefully minimal) changes to your code as the PR is up for discussion... still I would very much like to know if the proposed PR solves your problems you are seeing. Don't forget to instantiate in any child thread an instance of `stan::math::ChainableStack` before doing any autodiff. Doing so instantiates the AD tape... not doing so will lead to a segfault. Thanks!

riddell-stan:
@ahartikainen I just discovered this. https://bugs.python.org/issue35750 
The exit code -1073740940 is heap corruption. The comments in the linked 
bug report suggest we can use a debugger to find out what is going on. 
That could be really helpful.

riddell-stan:
@wds15 does your branch require c++14?

ahartikainen:
@riddell-stan I think that heap corruptions finally leads to the thread_local bug. I did some testing with PyStan with debugger, when I was trying to solve this problem.

wds15:
I don't think I need C++14, no. Just C++11 will do.

riddell-stan:
Looking at cmdstan's options, it looks like c++1y is turned on. This 
includes more than C++11, right?

wds15:
Yes. That's a bit more. I recall that the agreement was to use whatever gcc 4.9.3 supports which is C++11 and some/most/you better check of C++14...

riddell-stan:
@wds15 were any compiler flags added or changed between 2.18.1 and your pull request?

wds15:
I haven't changed any compiler flags, no. However, there are a few changes in develop vs 2.18.1, I think.

riddell-stan:
Important pieces of #151 have been added or are part of other PRs.

PR #153, opened 2019-03-05
WIP: stan math 1135 experiment (linux only)

riddell-stan:
updated source, hitting some problems getting models to compile. will try with cmdstan


riddell-stan:
On my local machine this segfaults when `import`ing the compiled extension module. This is likely `gdb` debug-able.

riddell-stan:
We'll try this again once 2.19 is officially released.

PR #175, opened 2019-07-30
activate conda base env first

ahartikainen:
Activates base env to set up paths correctly.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -14,7 +14,7 @@ install:
   - conda update -q conda
   - conda info -a
   - conda create -q -n test-env python=%PYTHON_VERSION%
-  - activate test-env
+  - conda activate test-env
   - conda install numpy cython scipy matplotlib pandas setuptools libpython m2w64-toolchain -c msys2 -q
   - gcc --version
   # Install the build dependencies of the project.
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,14 +1,14 @@
-os: Visual Studio 2015
+os: Visual Studio 2017
 platform: x64
 environment:
-  global:
-    MINICONDA: "C:\\Miniconda37-x64"
   matrix:
     - PYTHON_VERSION: 3.6
+      PYTHON_ROOT: "C:\\Miniconda36-x64"
     - PYTHON_VERSION: 3.7
+      PYTHON_ROOT: "C:\\Miniconda37-x64"
 
 install:
-  - set PATH=%MINICONDA%;%MINICONDA%\Scripts;C:\\msys64\usr\bin;%PATH%
+  - CALL "%PYTHON_ROOT%\\Scripts\\activate.bat"
   - conda activate base
   - conda config --set always_yes yes --set changeps1 no
   - conda update -q conda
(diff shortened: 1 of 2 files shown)


ahartikainen:
Ok, this is a bit annoying. I will continue this later.

ahartikainen:
This should be ready now.

.appveyor.yml
```diff
@@ -1,28 +1,49 @@
...
+  # purpose but it is problematic because it tends to cancel builds pushed
+  # directly to master instead of just PR builds (or the converse).
+  # credits: JuliaLang developers.
+  - ps: if ($env:APPVEYOR_PULL_REQUEST_NUMBER -and $env:APPVEYOR_BUILD_NUMBER -ne ((Invoke-RestMethod `
+      https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
+      Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
+        throw "There are newer queued builds for this pull request, failing early." }
+  
+  - SET PATH=%PYTHON_ROOT%;%PYTHON_ROOT%\Scripts;%PATH%
+  - CALL activate.bat base
```
riddell-stan: Can we remove the rollout thing? I don't think we need it.

Also, what does activate.bat do?

riddell-stan:
Maybe squash the commits? Any idea why it wasn't working before?

.appveyor.yml
ahartikainen: rollout? You mean %PYTHON_ROOT%? That is how cmd on windows handles env variables. We need to add them first, so activate.bat is on the PATH.

activate.bat adds necessary custom paths for conda, so it will work properly.

riddell-stan: I guess this was a conda bug or a conda/appveyor interaction bug? Is there a specific issue that we can monitor?

Trying to simplify things as much as possible.

ahartikainen:
Squash is fine for this.

Actually it is not yet "fixed", because the conda base for python 3.7 has somehow changed the file locations and currently needs some manual editing and even after that the conda env creation somehow fails. Not sure what is the correct fix for it, but I have not noticed those problems on my own computer or with conda base for python 3.6.

.travis.yml
```diff
@@ -1,6 +1,9 @@
 dist: xenial
 language: python
 python: 3.6
+branches:
```
riddell-stan: Why add this?

riddell-stan:
There are a lot of changes here, some to travis, some to appveyor, some to appveyor behavior that's independent of fixing the problem. Maybe split this up into three PRs?

.appveyor.yml
ahartikainen: https://github.com/conda/conda/issues/8865

https://github.com/conda/conda/issues/8836

.travis.yml
ahartikainen: this way, if you create a branch for httpstan and make a PR, tests are only run for the PR (now it runs duplicate amount for httpstan branches). After merge, it will run tests normally.

ahartikainen:
Yes, I can split this.

PR #176, opened 2019-08-02
Update source to Stan 2.20

### LINKED ISSUE #174: Release version with Stan 2.20
Stan 2.20 is out. This has some key threading improvements. Might even solve some Windows issues (fingers crossed).

riddell-stan:
Stan 2.20 is out.

Closes #174


riddell-stan:
Cython code needs to be updated to deal with this change "Add stan::model::model_base base class and and stan::model::model_crtp static adapter class for constructing models." https://github.com/stan-dev/stan/pull/2785

ahartikainen:
Do we need to (pre)compile these in setup.py?

riddell-stan:
I think you're right. I thought this didn't get in. I guess it did.

riddell-stan:
@ahartikainen looks like https://github.com/stan-dev/stan/pull/2785 is written in a way that's backwards compatible. If we want to benefit from faster compile time we need to change things. If we don't (yet), we don't have to change anything.

riddell-stan:
@ahartikainen This is ready.

(Also, looks like Windows compiles. We might want to see if it can sample as well without STAN_THREADS on.)

PR #177, opened 2019-08-02
Test master branch only

ahartikainen:
This PR enables branch-tests only for master.

This enables the developer to create a branch under `httpstan` and run tests only once for the PR.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,5 +1,6 @@
 os: Visual Studio 2015
 platform: x64
+skip_branch_with_pr: true
 environment:
   global:
     MINICONDA: "C:\\Miniconda37-x64"
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 dist: xenial
 language: python
 python: 3.6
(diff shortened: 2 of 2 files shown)


riddell-stan:
I'm a little confused about this one. Isn't the standard practice to have developers create branches on their own accounts? (I realize we didn't follow this in pystan always.)

riddell-stan:
I'd propose having all pull requests use branches from developers' accounts. This seems like a good idea to me because it creates a standard practice for both developers and other contributors (who don't have push access). It also seems like it's what astropy does https://github.com/astropy/astropy/branches . They seem to have a particularly clean contribution process.

ahartikainen:
That is totally fine. I think I will close this PR

On PyStan repo I have many times used the webpage interface for github to fix small problems, which causes a problem with fork (it can not be sync), So it has been just easier to create a new branch for the repo (and I have just kept one repo sync).

For httpstan and pystan-next we can define more strict rules for dev work. (Also I have terminal now with the android, so I can handle syncing with the phone).

PR #178, opened 2019-08-02
Manual rollout for appveyor

ahartikainen:
Skip unnecessary commit -tests for PR with multiple untested commits.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -8,6 +8,16 @@ environment:
     - PYTHON_VERSION: 3.7
 
 install:
+  # If there is a newer build queued for the same PR, cancel this one.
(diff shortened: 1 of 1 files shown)


ahartikainen:
I rebased this PR with #179 so the tests could be run correctly.

riddell-stan:
Can we see how bad things are with appveyor's "rollout builds" setting? I haven't had any problem so far. This adds a lot of confusing (to me) lines of code to the configuration.

ahartikainen:
Yes, I think we could actually start with that. We need to change settings in appveyor. Maybe if we are not happy with the results, let's add this manual hack.

ahartikainen:
I activated the rolling build option in appveyor (only for PR).

PR #179, opened 2019-08-02
Appveyor/update conda version

ahartikainen:
Change conda base from 3.7 to 3.6 in appveyor.  This is due to problems found in  https://github.com/conda/conda/issues/8865 and https://github.com/conda/conda/issues/8836

Updates also appveyor procedure to use mingw32-make from the m2w64-toolchain.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@ os: Visual Studio 2015
 platform: x64
 environment:
   global:
-    MINICONDA: "C:\\Miniconda37-x64"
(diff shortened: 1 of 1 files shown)


riddell-stan:
Thanks. Glad to have this by itself.

In the previous one you had ``os: Visual Studio 2017``. Did we want that?

ahartikainen:
I think there is no real gain with it.

ahartikainen:
Here is the list of pre-installed software on different images.  https://www.appveyor.com/docs/windows-images-software/

PR #181, opened 2019-08-06
fix: Remove (currently) unsupported sampling algo

### LINKED ISSUE #180: Support only a single sampling algorithm
There are two sampling algorithms right now. For the beta release the plan is to support a single default algorithm, hmc_nuts_diag_e_adapt. This is the default algorithm for cmdstan.

riddell-stan:
Only the default sampling algorithm, `hmc_nuts_diag_e_adapt` is
supported.

Also update a test to remove a reference to the unsupported algorithm.

Closes #180.


riddell-stan:
@ahartikainen Ready.

PR #186, opened 2019-08-07
Use .pyd for Windows modules

ahartikainen:
Fixes https://github.com/stan-dev/pystan-next/issues/81

This PR will use the correct `.pyd` extension for compiled modules on Windows. After this change, model loading from cache works on Windows.


riddell-stan:
Glad it fixes the problem. Maybe add something to the commit message so folks unfamiliar with this can understand why it fixes things? I hadn't heard of pyd before. https://docs.python.org/3/faq/windows.html#is-a-pyd-file-the-same-as-a-dll

We also need an actual test which exercises the caching functionality. Might be easy enough to verify that it takes less than half the time to compile a file a second time. Do you mind adding the caching test?

PR #195, opened 2019-08-09
fix: Restore STAN_THREADS

### LINKED ISSUE #191: Parallel sampling broken (STAN_THREADS related)
PR #176 intentionally introduced this bug. It needs to be fixed.

riddell-stan:
Restore use of STAN_THREADS which was removed to address a compatibility
issue.

This code is not expected to work using certain versions of gcc.

Closes #191

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,13 +5,7 @@ matrix:
   include:
     - python: 3.6
       env: TOXENV=py36
-      compiler: gcc
-    - python: 3.6
-      env: TOXENV=py36
       compiler: clang
-    - python: 3.7
-      env: TOXENV=py37
-      compiler: gcc
     - python: 3.7
       env: TOXENV=py37
       compiler: clang
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,16 +3,9 @@ language: python
 python: 3.6
 matrix:
   include:
-    - python: 3.6
-      env: TOXENV=py36
-      compiler: clang
-    - python: 3.7
-      env: TOXENV=py37
-      compiler: clang
     - env: TOXENV=flake8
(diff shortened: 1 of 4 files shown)


riddell-stan:
If things work with clang, we can remove gcc testing (or see if gcc 9 works).

ahartikainen:
Should we call `pytest` or even `python -m pytest` and use `-v`?

It is quite hard to say what failed without `-v`.

riddell-stan:
All tests should pass. This should be good to merge.

Tests pass for me locally with clang 6. On ubuntu 18.04 I have to run tests this way: ``CC="clang -stdlib=libstdc++" CCX="clang++ -stdlib=libstdc++" python3 -m pytest tests``

I've debugged the original problem extensively with gdb, see https://discourse.mc-stan.org/t/example-of-use-of-autodiffstacksingleton-stan-threads-related/10037 for discussion.

Short-term solution is to drop support for gcc for httpstan and pystan 3 on Linux. Clang works fine. (And maybe gcc will work on Windows?) With luck there will be a fix with Stan 2.21.

Thanks @wds15 for your help debugging this!

riddell-stan:
I think there's some sort of travis problem. Running tests via `tox` works fine locally for me, provided I export ``CC=clang CXX=clang++``.

ahartikainen:
I tested too with this branch on Ubuntu. (parallel bernoulli & pystan-next corr_matrix)

clang clang++ works, gcc g++ fails.

riddell-stan:
Let's see if a newer tox version helps.

riddell-stan:
@ahartikainen I managed to get travis working but I'll do that as part of a separate PR. I propose merging this as is.

PR #198, opened 2019-08-09
fix: Update tox version

riddell-stan:
This will not hurt and may help current continuous integration problems.


riddell-stan:
I'll go ahead and merge this as it doesn't touch code.

PR #202, opened 2019-08-10
fix(ci): Test using clang on travis

riddell-stan:
Do not merge yet.

Also, needs to be rebased on master once #195 is merged.

Either travis does not actually export CXX or there is some `libstdc++` issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,24 @@
-dist: xenial
+dist: bionic
 language: python
-python: 3.6
+# NOTE: when `language: python` is set, using `compiler: clang` or
+# `compiler: gcc` does nothing. These only work when `language: cpp`.
 matrix:
   include:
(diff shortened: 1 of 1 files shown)


ahartikainen:
Does python even use `CXX` kw?

It basically uses `CC` only. But for clang, clang++ is just clang with some extra settings.

riddell-stan:
Looks like distutils doesn't use CXX although some people want it to: https://bugs.python.org/issue1222585

I think the problem was that `compiler: clang` doesn't actually set CC to be clang -- which is a bit maddening since the travis docs say that it does (albeit on language: c++ projects). I guess people like us who are doing C++ extensions in Python aren't much of a priority.

riddell-stan:
This should be good to go now.

ahartikainen:
LGTM

PR #204, opened 2019-08-10
prefix to 'httpstan_' for tempfiles

ahartikainen:
Makes it easier to inspect tempfiles and also remove the old ones (Windows problem).

### DIFF
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -166,7 +166,7 @@ def import_model_extension_module(model_name: str, module_bytes: bytes):
     module_filename = f"{module_name}{'.so' if platform.system() != 'Windows' else '.pyd'}"
     assert isinstance(module_bytes, bytes)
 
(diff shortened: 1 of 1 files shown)


httpstan/models.py
```diff
@@ -248,7 +248,7 @@ def _redirect_stderr_to(stream: IO[Any]) -> int:
 
     # write files need for compilation in a temporary directory which will be
     # removed when this function exits.
-    with TemporaryDirectory() as temporary_directory:
+    with TemporaryDirectory('httpstan_') as temporary_directory:
```
riddell-stan: Maybe match the use above (i.e., with or without `prefix=`)?

(I'm assuming you'll have to change this anyway since black is going to complain.)

ahartikainen: Yes, there should be `prefix=`.

Issue #201, opened 2019-08-10
Test using clang on travis


riddell-stan:
PR #195 disabled testing on travis.

Getting clang to use a modern libstdc++ requires some fiddling.

ahartikainen:
Why `libc++` is not enough?

PR #207, opened 2019-08-10
fix(ci): Test osx using travis ci 

riddell-stan:
Needs to be rebased after #201 is merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,10 +11,37 @@ matrix:
     - python: 3.7
       env: TOXENV=coverage CC=clang
 
+    - os: osx
+      language: generic
+      env: TRAVIS_PYTHON_VERSION=3.6 TOXENV=py36 CC=gcc
+    - os: osx
+      language: generic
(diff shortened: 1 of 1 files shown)


riddell-stan:
Getting a "Fatal Python error: Bus error" with xcode 9. I'll see if it happens with xcode 10.

ahartikainen:
It is something to do with threading?

riddell-stan:
No threads are used. STAN_THREADS is not defined and we use processor-based parallelism.

ahartikainen:
```
httpstan/lib/stan/lib/stan_math/lib/boost_1.69.0/boost/type_traits/detail/has_binary_operator.hpp:54:156: error: no template named 'declval' in namespace 'std'; did you mean simply 'declval'?
```

This is `libc++ vs `libstd++` problem we have had before.

ahartikainen:
Would 10.4 work?

https://docs.travis-ci.com/user/languages/c/#c11c11-and-beyond-and-toolchain-versioning

riddell-stan:
Well, we gained Windows, lost mac, and lost gcc on Linux. Not a total loss. :)

PR #208, opened 2019-08-11
fix: Add typing information for httpstan

### LINKED ISSUE #129: Type annotations are not at 100%
All code should be type annotated. (We need to at least start measuring this.)

riddell-stan:
Closes #129


riddell-stan:
I'll merge this, given that (1) it's been 48 hours and (2) it doesn't change any code.

PR #211, opened 2019-08-11
fix: Remove unneeded compiler defines

### LINKED ISSUE #188: Boost defines are no longer needed
Apparently BOOST_RESULT_OF_USE_TR1 and BOOST_NO_DECLTYPE do not need to be defined with newer C++ compilers. This change should be made in pystan too.

Discussion: https://discourse.mc-stan.org/t/boost-defines/10087/1

riddell-stan:
Two compiler defines are no longer needed with C++11.

Closes #188


riddell-stan:
@ahartikainen have a look?

PR #216, opened 2019-10-06
feat: Use one-commit fork of math v2.20

riddell-stan:
This version reverts the new thread-local storage machinery.


riddell-stan:
Added a bit more detail to the commit message:

```
Specifically, the version of math present after the commit may be
described as the following:

- v2.20.0 (tag) with commit f39a3253d0e0d127975bc8ea37504f34da3fcab9
reverted by `git revert`. That's all.
```

riddell-stan:
I think dropping support for Windows makes sense. (So, back to where we were before.) We can get a 2.20 out that works on Linux and macOS.

ahartikainen:
I think it is fine for now.

riddell-stan:
@ahartikainen We need it to pass on Windows in order to build wheels (easily). Or should I just disable appveyor?

ahartikainen:
Can you just comment out the tests and let it just build the wheel?

riddell-stan:
@ahartikainen looks like I found a non-trivial test which works. Mind reviewing the PR? There should be no surprises here.

riddell-stan:
This is v2.20 with one commit reverted. There were other commits after the commit which "broke" everything. Those are still present. So STAN_THREADS should now work fine.

I'm thinking we can use this with pystan 2 as well. First I need to make a httpstan wheel release.

PR #218, opened 2019-10-11
fix: Remove dependency on uvloop

riddell-stan:
Do not install uvloop by default. Continue to use it if available.

Building uvloop sometimes fails on Linux. If Linux wheels for uvloop
become available, we might reconsider requiring it.


riddell-stan:
I'm going to merge this without review. No code is changed.

PR #223, opened 2019-10-20
feat: Replace 2.20 Stan source with 2.19 Stan source

riddell-stan:
Stan 2.20 created several problems:

- Building wheels
- Sampling broken on gcc on linux, gcc and clang on macOS

On the other hand, Windows appeared to work fine.

So this returns us to 2.19. Windows is not supported.


riddell-stan:
Skipped review. I'm going to skip the PR review step until we have a beta release of pystan 3. We need something stable before we can worry about making it unstable.

PR #224, opened 2019-10-21
chore: Update Stan source to tag v2.19.1

riddell-stan:
Update from 2.19.0 to 2.19.1


ahartikainen:
Do we wait for Stan 2.21 or still try to release Stan 2.20?

Do you want help with the 2.21 things (TBB)?

riddell-stan:
I definitely think we forget about 2.20 for pystan, httpstan, and pystan 3.

I think the next step is to get a version of httpstan which uses Stan 2.19.1 building without being header only. So we'd build the way cmdstan and rstan build. If we can do this, getting 2.21 working should be relatively easy. Does this make sense?

If so, we could potentially both try to figure this out. Do you want to try to get something working on PyStan 2 and I can try to get something working here?

PR #226, opened 2019-11-06
feat: Allow users to specify initial parameter values.

riddell-stan:


tests/test_user_inits.py
```diff
@@ -0,0 +1,103 @@
...
+    async def main() -> None:
+        random_seed = random.randrange(2 ** 16) + 1
+
+        sample_kwargs = {
+            "function": "stan::services::sample::hmc_nuts_diag_e_adapt",
+            "data": data,
+            "num_samples": 10,
+            "num_warmup": 0,
+            "random_seed": random_seed,
+        }
```
ahartikainen: This works without init?

riddell-stan: This isn't a very clear test. The idea is to provide no init and get some baseline. Starting with different inits (same random seed) should yield different results.

ahartikainen: Sounds reasonable

PR #227, opened 2019-11-06
ci: Update wheel downloading script for current builds [ci skip]

riddell-stan:


riddella:
Thanks. Yes, we should enable 3.8 everywhere if we're sure that appveyor and travis have been updated.

PR #228, opened 2019-11-07
ci: Add Python 3.8 testing, minor tox fixes

riddell-stan:


riddell-stan:
@ahartikainen want to have a look at the 3.8 related (conda?) error?

Would it be a good idea to see if we can't get things working on appveyor without conda (just python and pip)? Or is that a fool's errand?

ahartikainen:
I think we can skip 3.8 on Windows until conda has updated it's packages to support 3.8. There probably won't be 3.8 users, so nothing is "lost" and 3.8 is tested on *nix.

riddell-stan:
OK. Let's just do this from scratch once conda has packages ready.

ahartikainen:
conda-forge just updated pystan for 3.8 so I think this should work now.

Issue #229, opened 2019-11-07
Build and test requirements are not cleanly separated


riddell-stan:
`test-requirements.txt` contains Python packages which are only a requirement for building httpstan, not testing it. For example, the google package which installs protoc.

These should be separated. Doing so would also allow us to avoid mentioning specific package names in the travis and appveyor config -- which is certainly not a good practice (hidden configuration).

One might also take this opportunity to add pylint / autoflake to the checks.

riddell-stan:
This is really a bug. We don't specify the version of these build dependencies in the CI scripts but we do in test-requirements.txt.

PR #237, opened 2019-11-09
fix: Use Marshmallow 3.x for request validation

riddell-stan:
Three related packages updated:

- marshmallow
- webargs
- apispec

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,11 +5,11 @@
 pbr~=5.4 # Apache-2.0
 aiohttp~=3.6
 appdirs~=1.4
-webargs<6,>=5
-marshmallow<3.0.0rc6,>=3.0.0rc4
+webargs~=5.5
+marshmallow~=3.2
 numpy~=1.7
 Cython~=0.29
-apispec<1,>=0.21
+apispec[yaml,validation]~=3.1
 protobuf~=3.3
 setuptools>=28.8
 ujson~=1.3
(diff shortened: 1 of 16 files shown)


riddell-stan:
This still needs work. `apispec` changed its API a great deal.

riddell-stan:
This is ready to be looked at. Nothing of substance was changed. Most of the work involved cleaning up the OpenAPI spec (in the docstrings).

I suspect we'll be stuck depending on `marshmallow` as long as we want to automatically generate the OpenAPI spec (which is rather nice).

requirements.txt
```diff
@@ -5,11 +5,11 @@
...
 aiohttp~=3.6
 appdirs~=1.4
-webargs<6,>=5
-marshmallow<3.0.0rc6,>=3.0.0rc4
+webargs~=5.5
+marshmallow~=3.2
 numpy~=1.7
 Cython~=0.29
-apispec<1,>=0.21
+apispec[yaml,validation]~=3.1
```
ahartikainen: Could we do the same for our test / dev packages in `setup.py`

riddell-stan: Do you mean the `test-requirements.txt`? We don't define any packages in setup.py.

We do need to re-arrange what goes in requirements.txt and test-requirements.txt (and create, I think, a `build-requirements.txt`). #229 describes the issue a bit.

ahartikainen: Something like `extras_require=dict(test=get_requirements_dev()...)` <- from arviz https://github.com/arviz-devs/arviz/blob/master/setup.py

It is currently commented out, because arviz have github dependencies.

PR #245, opened 2019-11-11
refactor: Treat `data` the same as other args in services call

### LINKED ISSUE #230: `data` is treated differently from other arguments to services functions
This is a minor "code smell" issue. There's an explicit `data` argument in a couple of places where all other arguments are implicitly passed via kwargs. This should be fixed.

- [ ] No default values in the schema for call to `fit` endpoint (i.e., eliminate `missing={}`)
- [ ] explicit `data` argument should not appear anywhere (unless other arguments also mentioned explicitly)

riddell-stan:
In the call to a sample function in stan services, `data` is no
different than `init` or `num_samples`. Previously, it was treated as
different from other arguments. This commit corrects this.

Closes #230

### DIFF
--- a/httpstan/fits.py
+++ b/httpstan/fits.py
@@ -8,7 +8,7 @@ import httpstan
 import httpstan.stan
 
 
-def calculate_fit_name(function: str, model_name: str, data: dict, kwargs: dict) -> str:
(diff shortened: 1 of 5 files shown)


riddell-stan:
This needs #244 to be merged first and then rebased.

riddell-stan:
Should all work now.

PR #271, opened 2019-12-22
Serialize values using protocol buffers in Stan writer callback

riddell-stan:
Do not merge this. It works but it is currently slower than using JSON serialization.

The overall strategy, however, is sound. There are opportunities for making things faster here.

Other interfaces wishing to use protobuf for serialization of Stan draws may wish to take this as a starting point.


riddell-stan:
New approach which uses a shared library is faster.

PR #274, opened 2020-02-01
fix: Save error messages when initialization fails

### LINKED ISSUE #272: "Initialization failed" C++ exception not handled
Example code:

```
program_code = """
parameters {
  real y;
}
model {
  y ~ uniform(100, 101);
}
"""
```

There's an easy fix. The stan services call was not marked in Cython as potentially C++ exception-throwing.

Maybe more difficult: what do we return to the user via HTTP? Clearly not 200.

riddell-stan:
Previously (valuable) error messages were discarded when sampling
failed. This commit makes sure that the messages are saved.

Closes #272

### DIFF
--- a/tests/test_sampling_initialization_failed.py
+++ b/tests/test_sampling_initialization_failed.py
@@ -1,0 +1,53 @@
+"""Test sampling from a model where initialization will fail.
+
+Sampling from this model should generate ``Rejecting initial value`` logger
+messages followed by a C++ exception (which Cython turns into a Python
+exception). The exception is associated with the message ``Initialization
+failed.``
+"""
+import aiohttp
+import pytest
+
+import helpers
+
+program_code = """
+parameters {
+  real y;
+}
+model {
+  y ~ uniform(100, 101);
+}
+"""
+
+
+@pytest.mark.asyncio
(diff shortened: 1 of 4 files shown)


tests/test_sampling_initialization_failed.py
```diff
@@ -0,0 +1,54 @@
...
+}
+model {
+  y ~ uniform(100, 101);
+}
+"""
+
+@pytest.mark.asyncio
+async def test_sampling_initialization_failed(api_url: str) -> None:
+    """Test sampling from a model where initialization will fail."""
+    payload = {"function": "stan::services::sample::hmc_nuts_diag_e_adapt", "random_seed": 1}
```
ahartikainen: Will this fail always or is it just very probable (and should fail always due to seed)?

And could Stan3 changes affect this behaviour?

riddell-stan: Good question. It will always fail. By default, initial values are between -2 and 2 (sampled uniformly), see p. 57 in the CmdStan manual https://github.com/stan-dev/cmdstan/releases/download/v2.21.0/cmdstan-guide-2.21.0.pdf

I don't think stan3 will change anything.

PR #281, opened 2020-02-26
Switch build system to poetry

riddell-stan:
We cannot use pbr because we need to build and install shared libraries. The build system `poetry` allows us to do this. Using `setuptools` would also work but would require use of ``setup.py`` instead of the (more) declarative config of `poetry`.


riddell-stan:
@ahartikainen Have a look.

Once this gets in I'll add the new version of the queue_writer.hpp which serializes the messages in C++. It speeds things up a lot.

riddell-stan:
No updates to Stan. So we're still on 2.19.1.

I'm going to merge this. I am still a bit unhappy that poetry (with C extension modules) doesn't work with tox.

PR #286, opened 2020-03-03
feat: Encode Stan writer output using protobuf in C++

### LINKED ISSUE #102: httpstan is slow with small fits (benchmarks)
The benchmark here is 8 schools but with varying `J` to make things slow enough
that we can make measurements: 800 schools and 8000 schools. (The remaining
data is simply repeated the appropriate number of times.) In both cases the
models are compiled with `-DSTAN_THREADS`. Here are the results:

| J     | thin | cmdstan 2.18 | httpstan 0.6.1 | % slower |
| ----- | ---  | ------- | ---      | ---      |
| 800   | 1    |  4.5s   | 17s      | 270%     |  
| 8000  | 50   |  1m31s  | 1m45s    | 15%      |

TODO: add cmdstan without ``-DSTAN_THREADS``

Encoding things into protobuf format is the bottleneck. Since draws are coming out of Stan C++
much slower in the J = 8000 case, the bottleneck goes away. Two things would improve things, I suspect:

- Moving the protobuf encoding into Cython/C++. If the protobuf_writer emerges (/cc @sakrejda), this would solve the problem.
- Writing Stan C++ output to disk or to a cache in Cython rather than via the `async for` loop. Python loops are slow.

I'm happy with the progress so far. Things were far slower a week ago.

### LINKED ISSUE #282: Stan source does not need to be in repo
There's no reason to keep the Stan source in the repository. We can download the release tarball as part of the build process.

riddell-stan:
Previously Python was used. Doing everything directly in C++ is much
faster and we eliminate the dependency on the Python protobuf library.

~This version requires that the user install protobuf system-wide. Once
this solution is verified as working, we can remove this limitation by
copying the shared library into place.~

Before this is merged:
- [x] Clean up C++ to eliminate the use of stringstream, most uses are unnecessary.
- [x] Copy the protobuf library and include files into the package, use rpath (see @riddell-stan 's example repo)

This version is about 36% faster on longer fits (lots of draws).

Closes #282 
Closes #102


.travis.yml
```diff
@@ -19,6 +19,8 @@ cache: pip
 before_install:
   - python --version
   - if [[ -n $CC ]]; then $CC --version; fi
+  - sudo apt-get update
+  - sudo apt-get install -y libprotobuf-dev
 install:
   - pip install "poetry<2,>=1.0" tox
   - pip install "grpcio-tools~=1.8" "mypy-protobuf~=1.0"
```
riddell-stan: We now require protoc because we are generating C++ code. Eliminate references to grpcio-tools since it only supports Python.

riddell-stan:
This is not ready to be merged.

riddell-stan:
@ahartikainen How do I tell gcc on Windows about a library and include directories I have installed with `vcpkg`? Here, in .appveyor.yml, I've used ``vcpkg install protobuf protobuf:x64-windows`` but then gcc complains about not finding the header files for protobuf.

ahartikainen:
I think lib should on the PATH

riddell-stan:
This contains a lot of useful information: https://github.com/microsoft/vcpkg/issues/8577 . I'm still not sure precisely what to do. Testing is hard without a Windows computer. It seems like we're 95% of the way there.

httpstan/models.py
```diff
@@ -294,9 +296,10 @@ def _redirect_stderr_to(stream: IO[Any]) -> int:
         extension = setuptools.Extension(
             module_name,
             language="c++",
-            sources=[pyx_filepath.as_posix()],
+            sources=[pyx_filepath.as_posix(), callbacks_writer_pb_filepath.as_posix()],
```
riddell-stan: I'm not sure the callbacks_writer cpp file needs to be mentioned here.

riddell-stan:
@ahartikainen Have a look.

Everything important happens in the first commit, 4cd2ffffb593500837b66a4c679d153bc1278e0a. The file httpstan/callbacks_writer_parser.py is removed. The parsing work in that file is moved into C++ files, in httpstan/queue_logger.hpp and httpstan/queue_writer.hpp.

Encoding the protobuf messages in C++ means we need to link with the protobuf shared library, which we have to compile and place into the Python package (so subsequent model compilations can use it).

PR #287, opened 2020-03-04
 ci: Use Visual Studio 2019 instead of mingw

riddell-stan:


ahartikainen:
Does it work?

I think we still should let users use mingw?

I mean ~10gb for a compiler doesn't sound nice.

riddell-stan:
I was only thinking about testing. In theory, tests would go faster if we used VS since everything is already downloaded on appveyor.

PR #295, opened 2020-04-07
fix: Adjust gcc ABI compatibility settings

### LINKED ISSUE #294: httpstan 1.1.0 symbol mismatch
I'm getting this error:

```
{"status": "Bad Request", "message": "Failed to compile module: /tmp/httpstan_bm2p6kvl/model_8ab027191f.cpython-37m-x86_64-linux-gnu.so: undefined symbol: _ZNK6google8protobuf11MessageLite25InitializationErrorStringB5cxx11Ev", "code": 400}%                               
```

I checked the rpath setting in ``/tmp/httpstan_bm2p6kvl/model_8ab027191f.cpython-37m-x86_64-linux-gnu.so`` and it is looking in the right place. But the library ``libprotobuf-lite.so.22`` provides this symbol:

```
  1122: 00000000000ac4a8   103 FUNC    GLOBAL DEFAULT   11 _ZNK6google8protobuf11MessageLite25InitializationErrorStringEv
```

which is missing the "cxx11" of the previous one.

riddell-stan:
Closes #294

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,8 @@ build/protobuf-3.11.3:
 	curl --silent --location https://github.com/protocolbuffers/protobuf/releases/download/v3.11.3/protobuf-cpp-3.11.3.tar.gz | tar -C build -zxf -
 
 httpstan/include/google/protobuf httpstan/lib/libprotobuf-lite.so httpstan/bin/protoc: build/protobuf-3.11.3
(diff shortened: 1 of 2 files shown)


riddell-stan:
This needs a bit more work. Specifically, we need to set the ABI flag everywhere. libprotobuf needs to compile with it, for example.

riddell-stan:
That did the trick. Merging this since it's very minor.

PR #302, opened 2020-04-09
ci: CI with GitHub actions

ahartikainen:
This PR implements CI with GitHub actions.

Currently three environments are created, Linux, macOS and tox (Linux). These are easily extended to Windows, when we upgrade to 2.21+.


ahartikainen:
This is still a bit "work in progress". We need to check for correct settings so checks are run for PR. Also this PR is still missing all the uploads.

Feel free to edit the branch.

ahartikainen:
Ok, this runs fine on my forked repo, but it can not be run here due to use of a fork.

https://github.community/t5/GitHub-Actions/Github-Workflow-not-running-from-pull-request-from-forked/td-p/32979

I suggest we try Azure DevOps. The workflow is very similar.

riddell-stan:
Interesting. Shocking that Github Actions can't test pull requests.

Maybe we should collect the pros and cons of each service. travis-ci.com is working again for httpstan-wheels.

PR #303, opened 2020-04-09
ci: Test on macOS

riddell-stan:
Thanks to @ahartikainen for the macOS specific fix.


riddell-stan:
@ahartikainen Can I remove the `names` in the .travis.yml config or does codecov require them?

ahartikainen:
Yes you can. It can also be removed from the codecov command. It is just there to help debug things.

PR #306, opened 2020-04-10
fix: Give models from different `python`s distinct ids

### LINKED ISSUE #288: Use sys.executable in hash of model module filename
Models compiled by two different ``python``s should have distinct filenames. This is mainly useful for development.

(Double check that we also include the python version in the hash calculation.)

riddell-stan:
Previously, a model built under Python 3.8 would receive
the same id as a model built under Python 3.7.

Closes #288

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,7 +70,7 @@ program::
 
 This request will return a model name along with all the compiler output::
 
-    {"compiler_output": "", "name": "models/8ab027191f"}
+    {"name": "models/xc2pdjb4", "compiler_output": "In file included from …"}
 
 (The model ``name`` depends on the platform and the version of Stan.)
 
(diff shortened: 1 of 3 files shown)


riddell-stan:
This is not quite ready.

ahartikainen:
Should it include 64 / 32 bit info?

riddell-stan:
This is ready for review now.

Re: 32/64bit. This is only an issue for Windows, right? Will we need to worry about this in 2021? (I'm thinking PyStan 3 will be in beta for almost all of 2020.)

riddell-stan:
Added bit architecture. The docs recommend using sys.maxsize instead of platform.architecture().

README.rst
```diff
@@ -133,6 +133,9 @@ The next one is slightly more interesting. It reads::
      }
    }
 
+The Stan "fit", saved in ``myfit.bin``, aggregates all messages. By reading
+them one by one you can recover all messages sent by the Stan C++ library.
```
ahartikainen, line 137: This is good

PR #307, opened 2020-04-12
docs: Check that new release does not break pystan

### LINKED ISSUE #299: Add a "regression test" verifying httpstan does not break pystan3
Releasing a new version of httpstan should not break pystan3 / pystan-next, provided httpstan obeys semantic versioning.

In addition to running httpstan's tests, we should install pystan-next and run its tests. These both should pass.

riddell-stan:
Closes #299


ahartikainen:
Is there a way to do similar workflow for travis as we can do for Github Actions / Azure DevOps. This is something we could "easily" add to the workflow as a conditional step. (run and test, but don't fail etc)

riddell-stan:
I decided not to add the tests into the CI config because running the tests takes a really long time.

riddell-stan:
Closing this because GitHub made a merge commit when I clicked "Update branch". We should find a way to disable this button.

PR #308, opened 2020-04-12
test: Check that progress logger callback is being called

### LINKED ISSUE #277: Test for logger_callback missing
There is no test which verifies `logger_callback` is being called. (`logger_callback` updates the "% sampling complete" information available to clients.)

riddell-stan:
Closes #277

### DIFF
--- a/tests/test_logger_callback.py
+++ b/tests/test_logger_callback.py
@@ -1,0 +1,24 @@
+"""Test sampling from Bernoulli model."""
+import aiohttp
+import pytest
+
+import helpers
+
+program_code = "parameters {real y;} model {y ~ normal(0,1);}"
+
+
+@pytest.mark.asyncio
+async def test_logger_callback(api_url: str) -> None:
+    """Test that operation progress has been updated."""
+
+    payload = {"function": "stan::services::sample::hmc_nuts_diag_e_adapt"}
+    operation = await helpers.sample(api_url, program_code, payload)
+    operation_name = operation["name"]
+    operation_url = f"{api_url}/{operation_name}"
+    async with aiohttp.ClientSession() as session:
+        async with session.get(operation_url) as resp:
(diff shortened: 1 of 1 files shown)


tests/test_logger_callback.py
```diff
@@ -0,0 +1,25 @@
+"""Test sampling from Bernoulli model."""
+import aiohttp
+import pytest
+
+import helpers
+
+program_code = "parameters {real y;} model {y ~ normal(0,1);}"
```
ahartikainen, line 7: Should we import these simple models from somewhere (e.g `helpers.py`)

Also do we want to use cached versions or always compile new models?

tests/test_logger_callback.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+@pytest.mark.asyncio
+async def test_logger_callback(api_url: str) -> None:
+    """Test that operation progress has been updated."""
+
+    payload = {"function": "stan::services::sample::hmc_nuts_diag_e_adapt"}
+    operation = await helpers.sample(api_url, program_code, payload)
+    print(operation)
```
ahartikainen, line 16: is this debug line?

riddell-stan: Good catch. I need to remove that.

tests/test_logger_callback.py
riddell-stan: The cached models get used automatically when they exist. This seems to work well. We know that it compiled successfully (at least once).

riddell-stan: Putting the program_code strings for common models in helpers does sound like a good idea. Maybe in a separate PR though.

tests/test_logger_callback.py
riddell-stan: Removed.

PR #310, opened 2020-04-13
fix: Improve Exception messages

### LINKED ISSUE #275: Errors do not report the exception type
If something in the call to a services function throws an exception, the message is missing the exception type. The current code looks like this:

```
message, status = f"Error calling services function: `{exc}`", 400
```

This is a big problem when the message associated with `exc` is an empty string and the exception type carries all the information (e.g., an assertion error).

Adding some traceback information would also be welcome.

riddell-stan:
The previous code used `str(exc)` which does not display the Exception
type. The new message is much more informative. It also now includes
traceback information.

One additional test added.

Closes #275


httpstan/views.py
```diff
@@ -285,7 +286,10 @@ def _services_call_done(
         if exc:
             # e.g., "hmc_nuts_diag_e_adapt_wrapper() got an unexpected keyword argument, ..."
             # e.g., "Found negative dimension size in variable declaration"
-            message, status = f"Error calling services function: `{exc}`", 400
+            message, status = (
+                f"Exception during call to services function: `{repr(exc)}`, traceback: `{traceback.format_tb(exc.__traceback__)}`",
```
ahartikainen, line 290: How is the output from this combination?

riddell-stan: It doesn't add much information. The key thing is the exception type. Initialization failure is a ValueError but I think other errors should be different.

PR #317, opened 2020-04-22
fix: Drop test checking version number

riddell-stan:
The original test wasn't a terribly useful check.
A more automated way of updating the version would be welcome.


riddell-stan:
I'll delete the old 1.1.2 tag and re-tag.

ahartikainen:
So in optimal case, each tag could have an option to select what kind of release is done (e.g minor, major).

Is there a github app that could do that

riddell-stan:
There's a checklist in `doc/developers.rst`. Let's see if I can follow it next time :)

PR #318, opened 2020-04-24
ci: Use cache to avoid downloading tarballs

riddell-stan:
Downloading protobuf and Stan seem to cause build failures periodically.
These downloads can be cached.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,10 @@ matrix:
     - env: TOXENV=format
     - env: TOXENV=mypy
     - env: TOXENV=docs
-cache: pip
+cache:
+  - pip: true
+    directories:
+      - "$HOME/build/stan-dev/httpstan/build"
 before_install:
   - |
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -18,7 +18,10 @@ matrix:
     - env: TOXENV=format
     - env: TOXENV=mypy
     - env: TOXENV=docs
-cache: pip
+cache:
+  - pip: true
+    directories:
+      - "$HOME/build"
```
ahartikainen, line 24: Is this caching now also the compiled libs?

riddell-stan: Yes. I don't think this is harmful. There's a separate cache for each platform and language version. (Or at least that's what the travis-ci documentation leads me to believe.)

riddell-stan:
I think the directory needs to be:

```
/home/travis/build/stan-dev/httpstan/build
```

PR #327, opened 2020-04-25
feat: Add support for CVODES-requiring Stan models

### LINKED ISSUE #311: CVODES-related functionality unsupported
Now that we have a strategy for building and packaging shared libraries, it should not be difficult to resolve this.

riddell-stan:
This is the second attempt. The first attempt was reverted.

No code here deserves special comment. The Makefile rules for building
sundials are taken from CmdStan.

Closes #311

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 PYTHON ?= python
 PROTOBUF_FILES := httpstan/callbacks_writer_pb2.py httpstan/callbacks_writer.pb.cc
 STUB_FILES := httpstan/callbacks_writer_pb2.pyi
-LIBRARIES := httpstan/lib/libprotobuf-lite.so
+LIBRARIES := httpstan/lib/libprotobuf-lite.so httpstan/lib/libsundials_nvecserial.a httpstan/lib/libsundials_cvodes.a httpstan/lib/libsundials_idas.a
 INCLUDES := httpstan/include/google/protobuf httpstan/include/stan httpstan/include/stan/math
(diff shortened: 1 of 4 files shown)


riddell-stan:
The Makefile rules are challenging to get right. Someday we might explore a replacement. Right now the case for sticking with gnu make is overwhelming: we can copy rules directly from CmdStan.

Makefile
```diff
@@ -1,22 +1,21 @@
 PYTHON ?= python
 PROTOBUF_FILES := httpstan/callbacks_writer_pb2.py httpstan/callbacks_writer.pb.cc
 STUB_FILES := httpstan/callbacks_writer_pb2.pyi
-LIBRARIES := httpstan/lib/libprotobuf-lite.so
+LIBRARIES := httpstan/lib/libprotobuf-lite.so httpstan/lib/libsundials_nvecserial.a httpstan/lib/libsundials_cvodes.a httpstan/lib/libsundials_idas.a
```
ahartikainen, line 4: This was missing from the last try?

httpstan/models.py
```diff
@@ -304,7 +304,7 @@ def _redirect_stderr_to(stream: IO[Any]) -> int:
             define_macros=stan_macros,
             include_dirs=include_dirs,
             library_dirs=[f"{PACKAGE_DIR / 'lib'}"],
-            libraries=["protobuf-lite"],
+            libraries=["protobuf-lite", "sundials_cvodes", "sundials_idas", "sundials_nvecserial"],
```
ahartikainen, line 307: And this too?

Makefile
riddell-stan: I think this was what was missing.

PR #330, opened 2020-04-28
Update httpstan to use Stan 2.22.1

riddell-stan:
Most of the changes involve the Makefile.

This is currently segfaulting during sampling. Model compilation works fine.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,19 +1,19 @@
 PYTHON ?= python
 PROTOBUF_FILES := httpstan/callbacks_writer_pb2.py httpstan/callbacks_writer.pb.cc
 STUB_FILES := httpstan/callbacks_writer_pb2.pyi
-LIBRARIES := httpstan/lib/libprotobuf-lite.so httpstan/lib/libsundials_nvecserial.a httpstan/lib/libsundials_cvodes.a httpstan/lib/libsundials_idas.a
+LIBRARIES := httpstan/lib/libprotobuf-lite.so httpstan/lib/libsundials_nvecserial.a httpstan/lib/libsundials_cvodes.a httpstan/lib/libsundials_idas.a httpstan/lib/libsundials_kinsol.a
 INCLUDES := httpstan/include/google/protobuf httpstan/include/stan httpstan/include/stan/math
-INCLUDES_STAN_MATH_LIBS := httpstan/include/lib/boost_1.69.0 httpstan/include/lib/eigen_3.3.3 httpstan/include/lib/sundials_4.1.0
+INCLUDES_STAN_MATH_LIBS := httpstan/include/lib/boost_1.72.0 httpstan/include/lib/eigen_3.3.3 httpstan/include/lib/sundials_4.1.0 httpstan/include/lib/tbb_2019_U8
+MATH ?= build/math-3.1.1/
+TBB ?= $(MATH)lib/tbb_2019_U8
 
+default: all
 
-default: $(PROTOBUF_FILES) $(STUB_FILES) $(LIBRARIES) $(INCLUDES) $(INCLUDES_STAN_MATH_LIBS)
-
-
-build/protobuf-3.11.3/configure:
+build/protobuf-3.11.3:
 	@mkdir -p build
 	curl --silent --location https://github.com/protocolbuffers/protobuf/releases/download/v3.11.3/protobuf-cpp-3.11.3.tar.gz | tar -C build -zxf -
 
-httpstan/include/google/protobuf httpstan/lib/libprotobuf-lite.so httpstan/bin/protoc: build/protobuf-3.11.3/configure
+httpstan/include/google/protobuf httpstan/lib/libprotobuf-lite.so httpstan/bin/protoc: build/protobuf-3.11.3
 	@echo compiling with -D_GLIBCXX_USE_CXX11_ABI=0 for manylinux2014 wheel compatibility
 	cd build/protobuf-3.11.3 && ./configure --prefix="$(shell pwd)/httpstan" CXXFLAGS="-D_GLIBCXX_USE_CXX11_ABI=0" && make -j 8 install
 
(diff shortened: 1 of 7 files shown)


riddell-stan:
@wds15 Is there anything special I need to do with TBB  and Stan math if I am starting sampling within a thread? No initializing a special variable, right?

wds15:
No...the worst which can happen is that the tbb will use all cores on the machine.

Do you pre compile the services? If so, then one needs to be careful - we got this working for Rstan.

riddell-stan:
@wds15 thanks for the response. We do not currently pre-compile anything in Stan. We plan to do so, however.

I'll fire up gdb and try to figure out what is causing the segfault.

ahartikainen:
@wds15 so when we use threading (one process), will it use one pool for multiple chains?

We don't need to init tbb instance ourselves?

We sampling ends, what signals to kill the tbb?

wds15:
The TBB doc is really good:

https://software.intel.com/en-us/node/506056

You do not need to do anything.

What is a bit special in your case of httpstan is the handling of multiple threads per chain you sample.

I think it should be fine to call in each chain you sample first the `init_tbb_threadpool` function. This will create a static variable which holds the initialiser object for the tbb. Since the static variable is "global" within the translational unit of the chain everything will be fine - though it gets interesting how that works if you sample multiple chains in parallel for the same model. Maybe this is what causing a problem right now? In this case, I would start by not doing any TBB initialisation at all - the docs of the TBB say that this is ok; we just need to sort out how `STAN_NUM_THREADS` is being respected (and we would need to define what it actually means in your setting).

riddell-stan:
@wds15 Here's the (minimal) debugging information I'm getting:

```
Thread 10 "python3" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fffebfff700 (LWP 23748)]
stan::math::vari::operator new (nbytes=24) at ~/work/stan/httpstan/httpstan/include/stan/math/rev/core/vari.hpp:127
127         return ChainableStack::instance_->memalloc_.alloc(nbytes);
```

This comes in the call to the sampling function in stan services. Using math 3.1.1.

I'd welcome any tips on how to fix this or how to get more information out of the debugger. I've tried using tbb_debug and compiling with `-g`.

riddell-stan:
@wds15 Now I've tried adding ``init_tbb_threadpool`` before each call to the stan services function `hmc_nuts_diag_e_adapt`. Interestingly, it seems to work the first time. Subsequent calls yield the same segfault.

wds15:
You compile everything with STAN_THREADS?? Can you try to never call the init thing?

riddell-stan:
Yes. Everything compiled with STAN_THREADS. I tried it with and without 
the stan threadpool init function.

Without the init function it segfaults during (maybe after?) sampling 
always.

With the init function sampling works once and then segfaults on 
subsequent calls.

Does the difference tell you any useful information? Does the debugger 
message help?

ahartikainen:
Is there something destroys something, when sampling ends for a chain?

wds15:
How are the threads being created and managed?

Would it be possible to provide a dockerfile which sets this all up for easy debugging? Ideally a docker compose thing so that there are two containers...one with httpstan and the other to query it. 

I have an idea as to what goes wrong. Will explain when at a computer later (though that’s a guess).

ahartikainen:
Will this then currently work for 1 chain?

riddell-stan:
Yes, I can do this.

It might take some time, however. Right now I'm going to rewrite the 
plumbing of httpstan so we can use process-based parallelism. (I really 
just want something that works.) It appears that this works without any 
special changes.

wds15:
Docker would be great.

So here is how I understand what is going wrong here: For one you do create the threads in a library other than the TBB and for two you create programs at runtime such that these are distinct translational units. The combination of this is problematic. That is, in math it is assumed that you use TBB threads and the main thread is the only translational unit. In that design the so-called `ad_tape_observer` (see https://github.com/stan-dev/math/blob/develop/stan/math/rev/core/init_chainablestack.hpp) object is responsible for initialising the thread local AD tape. An instance of this thing is declared as global static variable - which makes this thing present in each translational unit. Now, it looks to me as if in the child threads which you create this `ad_tape_observer` object is not created (either never created or too late).

If I am right in my understanding here, then a solution could be to ensure that you have an instance of the `ad_tape_observer` instantiated before you call the Stan services.

However, what may still be a problem is that you end-up getting multiple statically declared `ad_tape_observer` instances - I am not sure on that for the moment.

A Docker thing would be good. I am not yet sure what is the best way to fix this. For now process based sampling avoids all of the above issues as you then do correctly instantiate the `ad_tape_observer` as it looks, but we should be able to support the thread-based thing as well.

wds15:
As I just came across a super odd thing with g++ (possibly a bug or an error on my side, not sure yet)... do you use g++? If so, can you switch to clang++ and try with that compiler? Maybe there is a simple fix in case switching to clang++ solves the matter.

(In case that's difficult for you I can prep a possible for fix for the issues at hand)

riddell-stan:
No luck. Same segmentation fault.

ahartikainen:
I tested this with your branch on Ubuntu (WSL2 on W10) and the tests pass (except cvodes)

CVodes error

    undefined symbol: CVodeGetSens

Are we missing some file for cvodes?

Do you run your tests on macOS?

ahartikainen:
```
make --version

GNU Make 4.1
Built for x86_64-pc-linux-gnu
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

```
gcc --version

gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

```
python --version

Python 3.7.3
```

```
cython (0.29.17)
numpy (1.18.4)
protobuf (3.11.3)
```

ahartikainen:
Also, I didn't define `STAN_NUM_THREADS`. Is that needed for the segfault?

riddell-stan:
I'm using a new approach that uses a ProcessPoolExecutor instead of a 
ThreadPoolExecutor. It's working. I'll update the draft pull request in 
a day or two.

The delay is due to having to rewrite the interprocess communication. 
The current approach passes data around using a thread-safe boost queue. 
But this queue can't be pickled by Python. To create a subprocess in 
Python basically everything you pass to the new process needs to be 
picklable.

riddell-stan:
I got things to work if we run the stan services sampling functions in different processes.

Since the switch from using a ThreadPoolExecutor to a ProcessPoolExecutor doesn't have anything to do with 2.23 per se, I'm going to create a different pull request for that. Once that's merged I will create a new PR for the move to 2.23.

Issue #336, opened 2020-05-12
make needs to be run twice


riddell-stan:
To properly (?) build libsundials shared libraries, `make` needs to be invoked twice.

The reason for this is, I believe, this line:

```
STAN_SUNDIALS_HEADERS := $(shell find $(MATH)stan -name *cvodes*.hpp) $(shell find $(MATH)stan -name *idas*.hpp)
```

The `find` command in the shell variable returns one thing when `make` is run the first time. It returns something else when `make` is run the second time.

I believe there exists a fix for this specific problem. See <https://stackoverflow.com/a/53760314>.

riddell-stan:
Fixed in df01dbebd3f37c32dc8e8bc476ed7541ab186849

PR #337, opened 2020-05-12
fix: Clean up Makefile rules

riddell-stan:
This makes the Makefile rules much better organized.  In particular, the httpstan Makefile now uses Stan Math's Makefile rules to build the shared libraries used by Stan Math. This relieves httpstan developers of the burden of maintaining complex Makefile rules.

Also adjusts the CI cache directory so builds are independent.

Closes #337 #336

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,82 @@
-PYTHON ?= python
+# Makefile for httpstan
+#
+# This Makefile generates code and builds the libraries used by httpstan
+#
+# Code generation rules appear in this Makefile. One httpstan-specific library
+# is built using rules defined in this Makefile. Other libraries used by all
+# Stan interfaces are built using rules defined in `Makefile.libraries`.  This
+# Makefile calls `make` to run `Makefile.libraries`. Note that some rules in
+# this Makefile copy libraries built by the other Makefile into their
+# httpstan-specific directories.
+
+PROTOBUF_VERSION := 3.11.3
+STAN_VERSION := 2.19.1
+MATH_VERSION := 2.19.1
+BOOST_VERSION := 1.69.0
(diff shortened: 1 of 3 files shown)


riddell-stan:
macOS seems like it is trying to build protobuf twice. Maybe `make` behaves differently?

riddell-stan:
I believe I found a way to directly use Stan Math's makefiles to build, say, libsundials. Using them will save us a lot of work here. It will also make it impossible to introduce errors in the library building process.

Makefile
```diff
@@ -27,117 +88,52 @@ httpstan/%_pb2.py: protos/%.proto httpstan/bin/protoc
...
-clean-sundials:
-	@echo '  cleaning sundials targets'
-	$(RM) $(wildcard $(sort $(SUNDIALS_CVODES) $(SUNDIALS_IDAS) $(SUNDIALS_NVECSERIAL) $(LIBSUNDIALS)))
-
+httpstan/lib/%: build/math-$(MATH_VERSION)/lib/sundials_$(SUNDIALS_VERSION)/lib/%
+	cp $< $@
 
 ###############################################################################
-# httpstan-specific libsundials
+# Build Stan-related shared libraries using Makefile.libraries
```
ahartikainen, line 131: Looks good. I don't fully grasp what is going on here, but if it works, then it works.

riddell-stan: I can improve the comment. `Makefile.libraries` is, essentially, a simplified version of stan math's makefile.

riddell-stan: Added more comments to explain that the new `Makefile.libraries` is a trimmed version of Stan Math's `makefile`.

PR #338, opened 2020-05-12
fix: Run stan::services functions in background using ProcessPool

riddell-stan:
Use a ProcessPoolExecutor instead of a ThreadPoolExecutor when
calling long-running stan::services functions.

Virtually all the changes in this commit follow from this change.
Switching to use a Unix domain socket to pass information from
subprocesses to the main process is required because the previous
method involved sharing memory between threads. Since we no longer
use threads, we cannot share memory in this way.

Remove the `PRAGMA journal_mode=WAL` setting for sqlite3 to avoid issues
associated with accessing the sqlite3 database from different processes.

Since we cannot use Windows currently due the use of Unix domain sockets,
remove the custom `TemporaryDirectory` in favor of the standard library version.


ahartikainen:
Why we can't use windows?

riddell-stan:
I'm going to raise the issue of Windows support for discussion on the Stan discourse forum once there's a clear path to Stan 2.23 on Linux and macOS.

We're getting close. I'll try to fix the `make` problems today or tomorrow.

riddell-stan:
This should be rebased after #337 is merged.

httpstan/cache.py
```diff
@@ -18,6 +18,11 @@
 logger = logging.getLogger("httpstan")
 
 
+def cache_filename() -> str:
+    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
```
ahartikainen, line 22: This is probably ok, but should this get \_\_version\_\_ from relative location?

```suggestion
    cache_path = appdirs.user_cache_dir("httpstan", version=.__version__)
```

riddell-stan: I think the verbose version is easier to read.

PR #342, opened 2020-05-16
feat: Update Stan to 2.22.1 and Stan Math to 3.1.1

riddell-stan:
This update is significant. Stan Math now requires the Intel TBB
library.

TODOs:
- [ ] Fix test failure 
- [ ] Create an issue on Stan Math about the Makefile error. Add the issue number to the Makefile.
- [ ] Rebase after #341 merged

### DIFF
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -259,14 +259,16 @@ def _build_extension_module(
             temporary_directory_path.as_posix(),
             os.path.join(httpstan_dir, "include"),
             os.path.join(httpstan_dir, "include", "lib", "eigen_3.3.3"),
-            os.path.join(httpstan_dir, "include", "lib", "boost_1.69.0"),
+            os.path.join(httpstan_dir, "include", "lib", "boost_1.72.0"),
             os.path.join(httpstan_dir, "include", "lib", "sundials_4.1.0", "include"),
+            os.path.join(httpstan_dir, "include", "lib", "tbb_2019_U8", "include"),
         ]
 
         stan_macros: List[Tuple[str, Optional[str]]] = [
             ("BOOST_DISABLE_ASSERTS", None),
             ("BOOST_PHOENIX_NO_VARIADIC_EXPRESSION", None),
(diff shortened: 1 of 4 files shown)


httpstan/models.py
```diff
@@ -278,14 +280,17 @@ def _redirect_stderr_to(stream: IO[Any]) -> int:
         # Note: `library_dirs` is only relevant for linking. It does not tell an extension
         # where to find shared libraries during execution. There are two ways for an
         # extension module to find shared libraries: LD_LIBRARY_PATH and rpath.
+        libraries = ["protobuf-lite", "sundials_cvodes", "sundials_idas", "sundials_nvecserial", "tbb"]
+        if platform.system() == "Darwin":
+            libraries.extend(["tbbmalloc", "tbbmalloc_proxy"])
```
ahartikainen, line 285: Why these are needed on macOS

riddell-stan: I don't know. Stan Math builds them for macOS and not for Linux. @wds15 do you want to comment?

riddell-stan:
Gcc works and clang doesn't. We didn't change *anything* except Stan and Math source code.

@wds15 any ideas? Remember we're only upgrading to 2.22.1 right now. (We could try 2.23 with ease.)

Moreover, normal sampling works fine. There are only a couple of failing tests (e.g., `tests/test_transformed_data_rng.py`).

I wonder if we missed some C++ change between 2.19.1 and 2.22.1.

ahartikainen:
are we missing `pthreads`?

httpstan/models.py
wds15: The tbbmalloc things are not needed on macOS, but they speed up Stan by 10-15%...even for single thread performance. Other platforms do not benefit that much or not at all.

riddell-stan:
Does it hurt to enable them on Linux? It would simplify the Makefile and 
the code here.

riddell-stan:
Closing as we'll go straight to 2.23.

PR #346, opened 2020-05-18
fix: Make the socket send/recv buffer larger

riddell-stan:
This may help with the macOS error about the buffer being exhausted.


riddell-stan:
Closing in favor of #347

PR #347, opened 2020-05-19
fix: Use stream instead of datagram over socket

riddell-stan:
Sending datagram messages (à la UDP) is extremely convenient. The code
was readable. Unfortunately, macOS has a very small send buffer. The
Stan logger and writer callbacks (which send us draws) send a lot of
messages and exceed the capacity of the buffer on macOS.

So we use the stream protocol instead. Stan will have to wait until we
are ready to send us draws.

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -9,6 +9,7 @@ Unix domain socket.
 import asyncio
 import concurrent.futures
 import functools
+import select
 import socket
 import sqlite3
 import tempfile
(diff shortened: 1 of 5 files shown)


httpstan/services_stub.py
```diff
@@ -80,27 +81,41 @@ async def call(
...
         future = asyncio.get_running_loop().run_in_executor(executor, lazy_function_wrapper_partial)
 
+        potential_readers = [socket_]
         while True:
-            try:
-                message = socket_.recv(8192)
-            except socket.timeout:
+            # note: timeout of 0 required to avoid blocking
+            readable, writeable, errored = select.select(potential_readers, [], [], 0)
+            for s in readable.copy():
```
riddell-stan, line 98: The `copy` isn't necessary.

httpstan/services_stub.py
```diff
@@ -80,27 +81,41 @@ async def call(
...
-                message = socket_.recv(8192)
-            except socket.timeout:
+            # note: timeout of 0 required to avoid blocking
+            readable, writeable, errored = select.select(potential_readers, [], [], 0)
+            for s in readable:
+                if s is socket_:
+                    conn, _ = s.accept()
+                    potential_readers.append(conn)
+                    continue
+                message = s.recv(1024 * 256)
```
ahartikainen, line 103: Is this some specific lenght or just convention?

httpstan/views.py
```diff
@@ -26,6 +26,10 @@
 logger = logging.getLogger("httpstan")
 
 
+# match a string such as `Iteration: 2000 / 2000 [100%]  (Sampling)`
+iteration_info_re = re.compile(rb"Iteration:\s+\d+ / \d+ \[\s*\d+%\]\s+\(\w+\)")
```
ahartikainen, line 30: Should there be `\s*` instead of real spaces in `\d+ / \d+ \[`?

Also `\w` or `\S`

httpstan/views.py
```diff
@@ -320,7 +324,8 @@ def logger_callback(operation: dict, message: bytes) -> None:
         # Deserializing it would be costly and require importing the protobuf Python module.
         if b"info:Iteration" not in message:
             return
-        operation["metadata"]["progress"] = message.split(b"info:", 1)[1].decode()
+        # When sampling completes rapidly, multiple iteration messages can be passed together. Return the final one.
+        operation["metadata"]["progress"] = iteration_info_re.findall(message).pop().decode()
```
ahartikainen, line 328: What happens if no progress info is given? Will this fail?
I think we don't have too long texts ro be parsed, so we don't need to optimize this. (reverse str and pattern, finditer+break (and reverse founds info))

httpstan/services_stub.py
riddell-stan: If it's too short we risk splitting messages. 4096 is what the docs recommend but this will split multiple sequences of logger messages if they appear all at once (which they do with very fast sampling runs). 1024 * 256 is the default send buffer size on Linux so I picked that.

httpstan/views.py
riddell-stan: I think this is correct. You can see how the messages are created in `src/stan/services/util/generate_transitions.hpp`:

```
      message << "Iteration: ";
      message << std::setw(it_print_width) << m + 1 + start << " / " << finish;
      message << " [" << std::setw(3)
              << static_cast<int>((100.0 * (start + m + 1)) / finish) << "%] ";
      message << (warmup ? " (Warmup)" : " (Sampling)");
```

httpstan/views.py
riddell-stan: We are guaranteed to have Iteration information due to the `if` statement on line 321.

PR #350, opened 2020-05-23
fix: Close sockets when an exception occurs during sampling

riddell-stan:
If a C++ exception occurs during the stan::services call (e.g.,
initialization failure) we need to close the sockets which send Stan
messages to httpstan. This is important because failing to close them
might mean there are lingering messages which are never received.  These
messages will likely contain information relevant to figuring out why
the exception occurred.


httpstan/anonymous_stan_model_services.pyx.template
```diff
@@ -140,39 +140,40 @@ def hmc_nuts_diag_e_adapt_wrapper(
...
+                init_buffer,
+                term_buffer,
+                window,
+                interrupt,
+                deref(logger),
+                deref(init_writer),
+                deref(sample_writer),
+                deref(diagnostic_writer),
+            )
+        finally:
```
ahartikainen, line 171: this is a clean solution

PR #355, opened 2020-06-02
fix: Require Python 3.8 on macOS

### LINKED ISSUE #340: Require Python 3.8 for macOS to avoid Catalina issues
If we require Python 3.8.2 or higher on macOS, we don't have to deal with the multiprocessing issue that seems to be circulating. See https://github.com/matplotlib/matplotlib/issues/15410#issuecomment-625027757

riddell-stan:
Resolves an issue with the multiprocessing module under macOS Catalina.

Closes #340

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,7 +32,10 @@ include = [
 ]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = [
+  {"version" = "^3.7", markers = "sys_platform != 'darwin'"},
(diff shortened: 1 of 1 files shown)


ahartikainen:
3.8 uses spawn instead of fork, right? Was that the problem?

riddell-stan:
The issue description links to the related matplotlib issue.

It looks like poetry doesn't let you do what I'm trying to do. macOS users will just have to work around the problem.

PR #356, opened 2020-06-02
fix: Do not use directories as make targets

### LINKED ISSUE #343: Directory targets in Makefile do not work
The directory targets (e.g., `build/math-$(MATH_VERSION)`) do not work because the timestamp on the directory changes whenever the contents of the directory change.

Here's the recommended trick, from <https://stackoverflow.com/a/3275488>:

```
I usually use a target of dir/.dirstamp instead of dir with a rule of mkdir -p dir && touch $@. The timestamps on dirs in unix change when the contents of the directory change preventing dependent rules from running. – docwhat Jan 31 '11 at 17:02
```

riddell-stan:
Directories should not be used as make targets because their
modification times change whenever their contents change.

Closes #343


riddell-stan:
`make` rebuilds things when the source has a modification date later than the target's. When we do an operation in a source directory (in some other rule), this updates the source directory's modification time.

Maybe we should use an order-only prerequisite instead, https://www.gnu.org/savannah-checkouts/gnu/make/manual/html_node/Prerequisite-Types.html I didn't know about this.

riddell-stan:
Closing in favor of #360

PR #358, opened 2020-06-03
fix: Pin Cython version to avoid surprises

riddell-stan:
Cython does has not always followed semantic versioning. By
allowing a narrower range of acceptable Cython versions we will
avoid unpleasant surprises.

Being conservative with versions is particularly important for
software using Stan since people will inevitably try to reproduce
results using very old versions of Stan.


riddell-stan:
No change is needed, if I'm reading the docs correctly. Docs: https://python-poetry.org/docs/dependency-specification/

PR #359, opened 2020-06-03
ci: Fix for macOS build timeout on travis

riddell-stan:
Compiling httpstan sometimes takes longer than 10 minutes on macOS.
Use `travis_wait` to extend timeout to 20 minutes.


riddell-stan:
I'm going to add a note to remove this when we start using stanc3. I'm fairly sure that compiling stanc3 takes a good chunk of the time.

PR #360, opened 2020-06-03
fix: Use order-only prerequisites for dirs in Makefile

### LINKED ISSUE #343: Directory targets in Makefile do not work

riddell-stan:
Using directories as targets in a Makefile creates problems.
We use order-only prerequisites to solve this problem.
The problem and the solution are discussed here:
https://stackoverflow.com/a/13418059

Closes #343

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,20 +38,19 @@ default: $(PROTOBUF_FILES) $(STUB_FILES) $(LIBRARIES) $(INCLUDES)
 ###############################################################################
 # Download archives via HTTP and extract them
 ###############################################################################
-
-$(PROTOBUF_ARCHIVE): build/archives
+build/archives:
 	@mkdir -p build/archives
-	@echo downloading archive $@
+
+$(PROTOBUF_ARCHIVE): | build/archives
+	@echo downloading $@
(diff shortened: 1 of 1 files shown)


riddell-stan:

Contributors should feel free to revert any commits in this pull request for any reason or for no reason at all. You can revert a commit with the `git revert` command.

riddell-stan:
@ahartikainen The above notice is my makeshift solution to the problem created by requiring reviews in a project which has less than 1 full-time developer. Any thoughts?

ahartikainen:
Yes, that sounds reasonable solution.

Good testing should handle most of the problems.

(I did read the PR when it came, but I have apparently failed to send the acception)

PR #361, opened 2020-06-03
chore: Update version

riddell-stan:
This will be the last version of httpstan which uses Stan 2.19.
1.1.3 differs from 1.1.2 in that it uses process-based parallelism
instead of thread-based parallelism. This may have performance
implications for models which can draw all their samples very quickly
(e.g., in a couple of seconds).


riddell-stan:
@ahartikainen after this merges I'm also going to squash the history of `master` so we remove the commits where we had the Stan and Math trees added to the repository. Removing those commits will speed things up. Math in particular is absolutely enormous (759M uncompressed).

PR #364, opened 2020-06-09
chore: Update version

riddell-stan:
Includes fix to include shared libraries in the wheel.


riddell-stan:
Merged without review as this PR qualifies as extremely minor.

PR #366, opened 2020-06-11
feat: Use stanc3 binary

### LINKED ISSUE #333: Switch to stanc3
I noticed from a mailing list message that stanc3 works with 2.22. httpstan can switch to that before moving to 2.23.

riddell-stan:
Download and install stanc ("stanc3") binary (written in OCaml).
Development of the C++ compiler has ceased.

Closes #333


httpstan/compile.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Returns:
+        str: C++ code
+
+    Raises:
+        ValueError: Syntax or semantic error in program code.
+
+    """
+    with tempfile.NamedTemporaryFile() as fh:
+        fh.write(program_code.encode())
```
ahartikainen, line 22: Is the default encoding utf-8?

Can Stan handle utf-16? (Japanese users for example might want to have this feature)

(In comments)

riddell-stan: I know there was a question about this at some point. In any event, this is an issue for the Stanc3 developers. The interfaces will not have any control over what encodings are acceptable by the compiler.

PR #367, opened 2020-06-11
fix: Include stanc in package

riddell-stan:
stanc is ignored in `.gitignore`, so poetry does not include it
unless instructed to.


ahartikainen:
Good catch, didn't notice that poetry uses `.gitignore`

Issue #365, opened 2020-06-09
CI does not test against an installed wheel


riddell-stan:
The current CI testing procedure tests httpstan in an odd way. It "installs" httpstan using something like the `pip install -e` / `python setup.py develop` method. This means we miss problems that can occur when the wheel is built, e.g., shared libraries failing to be included in the wheel itself.

Fixing this likely requires very small changes to the CI build instructions.

PR #368, opened 2020-06-11
fix: Use correct path to stanc binary

riddell-stan:
The full path to the stanc binary needs to be used in order to run the
stanc executable. Since the stanc binary could be installed anywhere, we
must retrieve the path using `importlib`.

This and previous pull requests are fixing bugs that are not detected by the current CI infrastructure. See #365

### DIFF
--- a/httpstan/compile.py
+++ b/httpstan/compile.py
@@ -1,5 +1,5 @@
 """Runs the `stanc` binary in a subprocess to compile a Stan program."""
-import os
+import importlib.resources
 import subprocess
 import tempfile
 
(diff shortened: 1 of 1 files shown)


httpstan/compile.py
```diff
@@ -18,17 +18,18 @@ def compile(program_code: str, stan_model_name: str) -> str:
...
-        fh.flush()
-        run_args = [
-            os.path.join("httpstan", "stanc"),
-            "--name",
-            stan_model_name,
-            "--print-cpp",
-            fh.name,
-        ]
-        completed_process = subprocess.run(run_args, capture_output=True, timeout=1)
+    with importlib.resources.path(__package__, "stanc") as stanc_binary:
```
ahartikainen, line 21: This is interesting solution

Normally I have done similar things like

    from pathlib import Path

    path = Path(httpstan.__file__).parent.resolve() / "stanc"

riddell-stan: The importlib approach doesn't assume that the library is on disk. It can, in theory, deal with packages that are encapsulated in a compressed file (e.g., zipapp).

PR #369, opened 2020-06-11
ci: CI with GitHub actions (transfer branch)

ahartikainen:
This PR implements CI with Github Actions.

- [x] tox tests
- [x] pytests with poetry
- [x] wheel: manylinux2014
- [x] wheel: macOS
- [ ] wheel tested after the build and before the upload
- [ ] new secrets: `PYPI_USERNAME` and `PYPI_PASSWORD`
- [ ] instructions how users need to activate Actions in their own fork.

First `tox` and `pytest` are run. Then if they succeed, wheels for manylinux2014 and macOS are build. Each build uploads wheels as an Artifact. If the pipeline is for a new tag, wheels are uploaded automatically to PyPI and to the release files.

Docs are built with `tox` and uploaded as an artifact.


ahartikainen:
This basically works, but this should be first either merged or send from the main repo.

Forks can not change workflow flags, so this PR will not trigger the action to run on my fork. After this is merged, forks sending PR would trigger actions to run on their own fork.

I suggest that I send this PR to another branch on `httpstan` and after that, we can work, clean, and finish the cloned PR. That would run Actions on httpstan repo and we could be certain that action files work as intended.

riddell-stan:
Looks interesting. I'm certainly interested in getting the wheel being built here rather than in a separate repo.

Can we perhaps focus on that first? Getting rid of httpstan-wheels and all that overhead would be a huge accomplishment.

As for the local/remote PR issue, I don't want to change the workflow. Everyone needs to submit PRs from their own branches, including us. Handling every contribution in the same way is important.

.github/workflows/workflow-httpstan.yml
```diff
@@ -0,0 +1,290 @@
...
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: dist/*.whl
+          tag: ${{ github.ref }}
+          overwrite: true
+          file_glob: true
+      - name: Upload to pypi
+        if: success() && startsWith(github.ref, 'refs/tags/v')
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
riddell-stan, line 225: Let's start with using the test PYPI repo. That way there's no risk of breaking things.

riddell-stan:
We also don't need username and passwords for pypi. We can generate an API token and use that. You should be able to generate a token.

.github/workflows/workflow-httpstan.yml
```diff
@@ -0,0 +1,290 @@
...
+        env:
+          CC: ${{ matrix.compiler }}
+        run: |
+          make
+      - name: Build wheel
+        env:
+          CC: ${{ matrix.compiler }}
+        run: |
+          python -m poetry run pip install --no-cache-dir Cython
+          python -m poetry build -v
```
riddell-stan, line 269: We need to run tests against the wheel before uploading it.

riddell-stan:
We can break up the workflows into different files, right? That seems like it might make things easier to read.

.github/workflows/workflow-httpstan.yml
ahartikainen: Yes, this is important.

I think we can create a new virtual env and install wheel on it.

ahartikainen:
We can break up them. That just divides the workflow to multiple items. Certainly we can test if that is confusing or not.

ahartikainen:
> As for the local/remote PR issue, I don't want to change the workflow. Everyone needs to submit PRs from their own branches, including us. Handling every contribution in the same way is important.

This local thing is only for this PR (fork can not activate actions use). Future PRs should work normally. (fork owners just need to enable github actions on their fork)

ahartikainen:
Ok, apparently we can not divide that to multiple files if we want to build conditionally.

I added a small header + some whitespace to separate different workflows.

I also added twine to upload test pypi for now.

riddell-stan:
I'd propose taking this slowly. Can we forget, for the moment, about the tox testing and normal testing and just do the wheel publishing to the test pypi? We can leave travis in place.

I have a different approach for regular testing, one which might allow us to have independent workflows (separate regular tests and wheel publishing (including tests) workflows).

.github/workflows/workflow-httpstan.yml
```diff
@@ -0,0 +1,341 @@
...
+    name: manylinux2014 wheel
+    needs: [pytest, tox-flake8, tox-format, tox-mypy, tox-docs]
+    runs-on: ${{ matrix.os }}
+    container:
+      image: quay.io/pypa/manylinux2014_x86_64:latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [cp37-cp37m, cp38-cp38]
+        compiler: [clang]
```
riddell-stan, line 178: Why are we using clang here?

.github/workflows/workflow-httpstan.yml
```diff
@@ -0,0 +1,341 @@
...
+        run: |
+          python -m poetry run pip install --no-cache-dir Cython
+          python -m poetry build -v
+      - name: Test wheel
+        env:
+          CC: ${{ matrix.compiler }}
+        if: success()
+        run: |
+          cd dist
+          python -m venv test_env --copies --clear
```
riddell-stan, line 310: Are --copies and --clear necessary?

ahartikainen:
Yes.

What if we do this in a `httpstan` branch, I can leave my branch as an example.

The reason for making this in `httpstan` and not in a fork, is that we can test the workflow without any hacks. (I currently have one PR for my fork, so I can see what is going on).

.github/workflows/workflow-httpstan.yml
ahartikainen: Probably not, I copied this from my other repo

.github/workflows/workflow-httpstan.yml
ahartikainen: I just needed to update the compiler and used only clang.

The gcc update is similar.

Which one should we use (for manylinux2014)?

riddell-stan: I think we should match everything to the configuration we use in httpstan-wheels. We use manylinux2014 with few changes. I don't think we change the compiler at all.

riddell-stan:
I don't mind merging a PR which just does the wheel uploading (to Test PyPI). It doesn't bother me if it hasn't been fully tested.

riddell-stan:
BTW, #373 shows how you can get the test dependencies out of poetry. This will likely be useful for testing the wheels.

.github/workflows/workflow-httpstan.yml
ahartikainen: We don't but I think multibuild does. Because the default compiler is 4.x?

Do we need C++14 for math?

I would have used multibuild, but I think it currently does not support Actions (there is an open issue, and I don't know if we could implement that actions setup for it).

Action tools need to wrap the tool with Docker or JS.

The thing I'm not sure / happy is that the procedure for macOS wheel might not be compatible with older macOS. Not sure how multibuild handles that.

ahartikainen:
Ok, I will close this one and create a new PR which uploads wheels to test pypi when a tag is created.

btw, I added PYPI_USERNAME, PYPI_PASSWORD and TEST_PYPI_PASSWORD to httpstan secrects. These are based on the token procedure.

What do you think about the uploading the wheel to tag files (in github)? This can overwrite the old wheel even when pypi doens't support it.

What about the artifact upload, which uploads wheel everytime we want (e.g. when master is pushed)? So we can install the master with wheel if needed. These files are automatically deleted after a period (30 days?) and are mostly for dev work. (Like uploading compiled docs).

riddell-stan:
Let's take one step at a time? We can first replace the functionality of 
httpstan-wheels: Build, test, and upload the wheel to (Test) PyPI on 
tagged commits.

If everything works we can consider other uses of GitHub Actions.

PR #373, opened 2020-06-12
ci: Build and install wheel, then run tests

### LINKED ISSUE #365: CI does not test against an installed wheel

riddell-stan:
Previously we tested in an environment which might be called poetry's
"developer mode". Now we build a wheel, install the wheel, and test.
Running tests in this sort of environment is better because the
environment resembles the one users will have.

Closes #365

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,9 +45,12 @@ install:
   - make  # generate Python protobuf files and build shared libraries
 script:
   - tox
-  - poetry install -v
+  - poetry build -v
+  # export test dependencies from pyproject.toml, install them
+  - poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
+  - python -m pip install dist/*.whl
   # Running pytest with `--forked` runs each test in a separate subprocess. Otherwise flaky tests
   # pass when this is used. These flaky tests only appear when `clang` is used. With `gcc` all tests pass.
-  - poetry run pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS tests
+  - mkdir testdir && cd testdir && python -m pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
 after_success:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -45,9 +45,12 @@ install:
   - make  # generate Python protobuf files and build shared libraries
 script:
   - tox
-  - poetry install -v
+  - poetry build -v
+  # export test dependencies from pyproject.toml, install them
+  - poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
```
ahartikainen, line 50: Is this making the venv more isolated?

.travis.yml
```diff
@@ -45,9 +45,12 @@ install:
...
+  - pip install dist/*.whl
```
ahartikainen, line 51: will this install wheel locally? Will the test be run against this or poetry set up?

riddell-stan: Good point. I need to drop the `poetry run` bit.

.travis.yml
riddell-stan: What venv?

ahartikainen: When we first build the wheel with poetry, poetry handles the requirements for build (poetry creates a venv for this, right?).

Then we install runtime requirements with this line, again for the venv? Or is that `--dev` line doing something I don't understand.

After this we install the wheel to regular python, which is clean in our workflow so testing against it is valid way to do the testing.

ahartikainen: Oh, now I get it.

We don't have requirements.txt file, but poetry creates one with the help of `pyproject.toml`

And `--dev` exports [tool.poetry.dev-dependencies]

riddell-stan: That's right.

PR #374, opened 2020-06-12
Upload wheels to PyPI

ahartikainen:
This PR builds httpstan wheels with Github Actions when any tag is created.

- [x] wheel: manylinux2014
- [x] wheel: macOS
- [x] new secrets: `PYPI_PASSWORD` and `TEST_PYPI_PASSWORD` (done in github)

### DIFF
--- a/.github/workflows/wheel_macOS.yml
+++ b/.github/workflows/wheel_macOS.yml
@@ -1,0 +1,63 @@
+name: wheel - macOS
+
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  wheel_macos:
+    name: macOS wheel
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-10.15]
+        python-version: [3.7, 3.8]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: check C++ compiler and Python
+        run: |
+          clang --version
+          python --version
+      - name: Install dependencies
+        run: |
+          python -m pip install --no-cache-dir --upgrade pip
+          python -m pip install --no-cache-dir "poetry<2,>=1.0"
+          python -m pip install --no-cache-dir "mypy-protobuf~=1.0"
+      - name: Build libraries
+        run: |
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        run: |
+          python -m pip install wheel
+          python -m pip install delocate
+          delocate-wheel -v dist/*.whl -w wheelhouse
+      - name: Install wheel
+        run: |
+          python -m pip install wheelhouse/*.whl
+      - name: Install test dependencies
+        if: success()
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt
+          python -m pip install -r requirements.txt
+      - name: Tests
+        if: success()
+        run: |
+          mkdir testdir
+          cd testdir
+          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 ../tests
+          cd ..
+      - name: Upload to pypi
+        env:
+          TWINE_TOKEN: ${{ secrets.TEST_PYPI_PASSWORD }}
+        if: success()
+        run: |
+          python -m pip install twine
+          python -m twine upload --repository-url https://test.pypi.org/legacy/ -u __token__ -p ${TWINE_TOKEN} --skip-existing wheelhouse/*
--- a/.github/workflows/wheel_manylinux.yml
+++ b/.github/workflows/wheel_manylinux.yml
@@ -1,0 +1,62 @@
+name: wheel - manylinux2014
+
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  wheel_manylinux:
+    name: manylinux2014
+    runs-on: ${{ matrix.os }}
+    container:
+      image: quay.io/pypa/manylinux2014_x86_64:latest
+    strategy:
+      matrix:
+        os: [ubuntu-18.04]
+        python-version: [cp37-cp37m, cp38-cp38]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        run: |
+          echo "::add-path::/opt/python/${{ matrix.python-version }}/bin/"
+      - name: check C++ compiler and Python
+        run: |
+          gcc --version
+          python --version
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade --no-cache-dir pip
+          python -m pip install --no-cache-dir "poetry<2,>=1.0"
+          python -m pip install --no-cache-dir "mypy-protobuf~=1.0"
+      - name: Build libraries
+        run: |
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        run: |
+          auditwheel repair dist/*.whl -w wheelhouse
+      - name: Install wheel
+        if: success()
+        run: |
+          python -m pip install wheelhouse/*.whl
+      - name: Install test dependencies
+        if: success()
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt
+          python -m pip install -r requirements.txt
+      - name: Tests
+        if: success()
+        run: |
+          mkdir testdir
+          cd testdir
+          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 ../tests
+      - name: Upload to pypi
+        env:
+          TWINE_TOKEN: ${{ secrets.TEST_PYPI_PASSWORD }}
+        if: success()
+        run: |
+          python -m pip install twine
+          python -m twine upload --repository-url https://test.pypi.org/legacy/ -u __token__ -p ${TWINE_TOKEN} --skip-existing wheelhouse/*


.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,57 @@
...
+        if: success()
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
+          mkdir testdir && cd testdir
+          poetry run pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
+          cd ..
+      - name: Upload to pypi
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
riddell-stan, line 52: No need for PyPI username, we can use project-specific, scoped API tokens, https://pypi.org/help/#apitoken

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
+          mkdir testdir && cd testdir
+          poetry run pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
+          cd ..
+      - name: Upload to pypi
+        if: success()
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
riddell-stan, line 62: See above about using API tokens.

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,66 @@
...
+          python -m poetry build -v
+      - name: Fix wheel
+        env:
+          PLAT: manylinux2014_x86_64
+        run: |
+          for whl in dist/*.whl; do
+            auditwheel repair "${whl}" --plat "${PLAT}" -w "./dist"
+            rm ${whl}
+          done
+          python -m pip install  --no-cache-dir dist/*.whl
```
riddell-stan, line 48:
```suggestion
          python -m pip install  dist/*.whl
```

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        env:
+          PLAT: manylinux2014_x86_64
+        run: |
+          for whl in dist/*.whl; do
```
riddell-stan, line 44: Is the for loop necessary?

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,57 @@
...
+          python -m pip install --no-cache-dir dist/*.whl
+      - name: Tests
+        env:
+          PYTEST_ARGS: "--forked"
+        if: success()
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
+          mkdir testdir && cd testdir
+          poetry run pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
```
riddell-stan, line 48:
```suggestion
          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
```

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,66 @@
...
+          python -m pip install  --no-cache-dir dist/*.whl
+      - name: Tests
+        env:
+          PYTEST_ARGS: "--forked"
+        if: success()
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
+          mkdir testdir && cd testdir
+          poetry run pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
```
riddell-stan, line 57:
```suggestion
          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
```

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,66 @@
...
+          cd ..
+      - name: Upload to pypi
+        if: success()
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.TEST_PYPI_PASSWORD }}
+        run: |
+          python -m pip install --no-cache-dir twine
```
riddell-stan, line 65:
```suggestion
          python -m pip install twine
```

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,57 @@
...
+          cd ..
+      - name: Upload to pypi
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          TWINE_PASSWORD: ${{ secrets.TEST_PYPI_PASSWORD }}
+        if: success()
+        run: |
+          python -m pip install --no-cache-dir twine
```
riddell-stan, line 56:
```suggestion
          python -m pip install twine
```

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,57 @@
...
+        os: [macos-latest]
+        python-version: [3.7, 3.8]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: check C++ compiler and Python
+        run: |
```
riddell-stan, line 22: I think clang --version and python --version are enough. That's two lines instead of five.

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,66 @@
...
+        os: [ubuntu-latest]
+        python-version: [cp37-cp37m, cp38-cp38]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        run: |
+          echo "::add-path::/opt/python/${{ matrix.python-version }}/bin/"
+      - name: check C++ compiler and Python
+        run: |
+          which gcc
```
riddell-stan, line 24: gcc --version and python --version are enough.

.github/workflows/wheel_macOS.yml
ahartikainen: I'm doing it, it is just under the secrects.PYPI_USERNAME

.github/workflows/wheel_manylinux.yml
ahartikainen: actually, no (I didn't know we could pip install with wildcards)

.github/workflows/wheel_macOS.yml
ahartikainen: I can remove the other lines.

Usually, I might use some special compiler/python so printing the which is an old habit (you don't believe how often people run tests with the wrong python)

.github/workflows/wheel_manylinux.yml
ahartikainen: Oh, this is not the install line. I don't know if auditwheel can take wildcards

ahartikainen: It does. On the other hand, this loop also removes the old wheel.

Should we move fixed wheel to another location or just keep the current workflow?

.github/workflows/wheel_macOS.yml
riddell-stan: Twine username can be a string `__token__`, right? That means we don't need a variable. Or have I misread the instructions?

.github/workflows/wheel_manylinux.yml
riddell-stan: `multibuild`'s approach is interesting. It puts the repaired wheel in a new directory, avoiding the need to delete the old one. The new directory is called `wheelhouse`.

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,71 @@
...
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        env:
+          PLAT: manylinux2014_x86_64
+        run: |
+          for whl in dist/*.whl; do
+            auditwheel repair "${whl}" --plat "${PLAT}" -w "./dist"
```
riddell-stan, line 42: Looking at the multibuild and auditwheel code, I don't see them using ``--plat``. Perhaps they autodetect the platform (which would make sense).

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,61 @@
...
+  push:
+    tags:
+      - '*'
+jobs:
+  wheel_macos:
+    name: macOS wheel
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest]
```
riddell-stan, line 13: Should we perhaps fix the macos version to mirror the version we are using in `httpstan-wheels`. I think this would be a good idea because changes in macOS versions have been known to break everything (e.g., Catalina).

.github/workflows/wheel_macOS.yml
ahartikainen: Yes, I just put it under secrets, we can pull it out and use bare `__token__`

.github/workflows/wheel_manylinux.yml
ahartikainen: I think that plat is somewhere. I can try without it. Maybe the docker has already that as a default?

.github/workflows/wheel_macOS.yml
ahartikainen: Is there a need to build for different macOS versions? It used to be easy, now macOS binary/compilation stuff is small mystery for me.

.github/workflows/wheel_macOS.yml
riddell-stan: It's not conventional to make a variable for a constant string value, is it? Or am I missing something?

.github/workflows/wheel_manylinux.yml
riddell-stan: I'm happy to change things from our current setup (see httpstan-wheels) if there's a good reason to do so. Otherwise I'd prefer to keep things the same. I barely understand all the details here so I think it's best to stick to what we know works (i.e., httpstan-wheels)

.github/workflows/wheel_macOS.yml
ahartikainen: maybe not

.github/workflows/wheel_macOS.yml
riddell-stan: Can we pick a specific macos version? That's what we do with httpstan-wheels, we use the travis tag "xcode11.3" which uses macOS 10.14.6 (see https://docs.travis-ci.com/user/reference/osx ). Could we perhaps start with the equivalent to that? If you want to change things after we get things working, we could have a new PR.

riddell-stan:
Is the macOS build missing a step? In multibuild the `delocate` program gets run. I think it's similar to the ``auditwheel repair`` step.

.github/workflows/wheel_manylinux.yml
riddell-stan: Your approach is ok. I don't like the for loop.

Maybe just output the wheel into the base directory (i.e., not dist). That would work, right? And it would eliminate a few lines of code.

ahartikainen:
Yes!, it is my missing step (I was wondering that I'm missing something there)

.github/workflows/wheel_manylinux.yml
riddell-stan: I was thinking this was a minor issue but now I'm not so sure. I think we should be explicit here and upload the platform-specific wheel by name. We know what the filenames look like. For linux the filename looks like this: `httpstan-*-manylinux*_x86_64`. For macOS it looks like this: `httpstan-*-macosx_*_intel.whl`.

The reason for this is the following -- we want to avoid accidentally uploading the un-repaired "any" wheel. This mistake just happened in httpstan-wheels with 2.0.0. You can see on PyPI that a "any" whl was uploaded. It probably has linux specific shared libraries in it. So it's broken on macOS.

.github/workflows/wheel_macOS.yml
ahartikainen: changed to `__token__`

.github/workflows/wheel_manylinux.yml
ahartikainen: I put the fixed wheel to a local folder called `wheelhouse`, and removed the loop.

(`/wheelhouse` is not a writable location with macOS image)

.github/workflows/wheel_macOS.yml
ahartikainen: I'm not sure if possible

https://github.blog/changelog/2019-11-06-github-actions-macos-virtual-environment-updated-to-catalina/

Is there possibility for docker?

riddell-stan: Let's just use `macos-10.15` rather than macos-latest so we know which image we are using and the image used doesn't change unexpectedly when github adds 10.16

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        env:
+          PLAT: manylinux2014_x86_64
+        run: |
+          auditwheel repair dist/*.whl --plat "${PLAT}" -w "./wheelhouse"
```
riddell-stan, line 41:
```suggestion
          auditwheel repair dist/*.whl -w wheelhouse
```
I'm still a bit confused here. Is the wheelhouse directory already created? If not we need to create it.

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        run: |
+          python -m pip install wheel
+          python -m pip install delocate
+          delocate-wheel -v dist/*.whl -w ./wheelhouse
```
riddell-stan, line 40:
```suggestion
          delocate-wheel -v dist/*.whl -w wheelhouse
```

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,66 @@
...
+  push:
+    tags:
+      - '*'
+jobs:
+  wheel_macos:
+    name: macOS wheel
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [macos-latest]
```
riddell-stan, line 13:
```suggestion
        os: [macos-10-.15]
```

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,67 @@
...
+      image: quay.io/pypa/manylinux2014_x86_64:latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        python-version: [cp37-cp37m, cp38-cp38]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        run: |
+          echo "::add-path::/opt/python/${{ matrix.python-version }}/bin/"
```
riddell-stan, line 21: The python version doesn't get configured automatically?

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,66 @@
...
+          make
+      - name: Build wheel
+        run: |
+          python -m poetry build -v
+      - name: Fix wheel
+        run: |
+          python -m pip install wheel
+          python -m pip install delocate
+          delocate-wheel -v dist/*.whl -w ./wheelhouse
+          delocate-addplat -x 10_9_intel -x 10_10_intel .wheelhouse/*.whl
```
riddell-stan, line 41: I'd like to double check that these are the delocate command we currently run in httpstan-wheels. I think it's important that we first reproduce what we are doing currently (since we know it works). Once it's working we can experiment with changes to the procedure.

.github/workflows/wheel_manylinux.yml
riddell-stan: Sorry, I see that -w takes a directory as an argument. It creates it if it doesn't exist I gather?

.github/workflows/wheel_macOS.yml
riddell-stan, line 41:
```suggestion
          delocate-addplat -x 10_9_intel -x 10_10_intel wheelhouse/*.whl
```
looks like there was an extra period

riddell-stan: I don't believe we use delocate-addplat in httpstan-wheels right now. Did you have a reason for adding it?

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,67 @@
...
+      - '*'
+jobs:
+  wheel_manylinux:
+    name: manylinux2014
+    runs-on: ${{ matrix.os }}
+    container:
+      image: quay.io/pypa/manylinux2014_x86_64:latest
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
```
riddell-stan, line 15:
```suggestion
        os: [ubuntu-18.04]
```

We currently use ubuntu 18.04 in httpstan-wheels

.github/workflows/wheel_macOS.yml
riddell-stan: I believe `delocate-wheel` is the only command which is required.

.github/workflows/wheel_manylinux.yml
```diff
@@ -0,0 +1,67 @@
...
+          python -m poetry export -f requirements.txt --dev -o requirements.txt
+          python -m pip install -r requirements.txt
+      - name: Tests
+        env:
+          PYTEST_ARGS: "--forked"
+        if: success()
+        run: |
+          mkdir testdir
+          cd testdir
+          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 $PYTEST_ARGS ../tests
```
riddell-stan, line 59: How does `success()` get determined? Does it look at the last command? If so, `cd ..` is likely to exit successfully.

riddell-stan:
fyi, ``--forked``` is no longer needed, see #384

ahartikainen:
Great.

I will update the branch.

.github/workflows/wheel_manylinux.yml
ahartikainen: Yes

.github/workflows/wheel_manylinux.yml
ahartikainen: Not on the manylinux image

.github/workflows/wheel_manylinux.yml
ahartikainen: 18.04 is just the linux that runs the manylinux2014 image.

riddell-stan:
Perhaps we should split these up into macos and linux. It seems like the issues are a little different.

On Linux, one difference here (vs. httpstan-wheels / multibuild) is that with multibuild the wheel is built using the manylinux container but then is installed (and tested) in an entirely different environment (i.e., the normal ubuntu environment). Can we replicate this? If not, we should probably stick with multibuild since it's likely to be much more robust as its testing the manylinux wheel in a realistic environment.

ahartikainen:
It runs all the steps on multibuild image. The docker is run on another linux.

.github/workflows/wheel_macOS.yml
```diff
@@ -0,0 +1,63 @@
...
+          # export test dependencies from pyproject.toml, install them
+          python -m poetry export -f requirements.txt --dev -o requirements.txt
+          python -m pip install -r requirements.txt
+      - name: Tests
+        if: success()
+        run: |
+          mkdir testdir
+          cd testdir
+          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 ../tests
+          cd ..
```
riddell-stan, line 56: I'm still a bit confused about `success()` here. Is it determined by all commands exiting with exit code 0? Or just by the last one. If it's the latter, then `cd ..` will certainly succeed even if the tests fail.

We ran into this problem earlier on travis where the tests could fail but because codecov succeeded, the build was labeled successful by travis.

ahartikainen: We can try what happens

riddell-stan: I'd prefer to consult the documentation than guess-and-check.

It does look like github actions runs bash with `set -e -o pipefail`, so this command should be safe. See https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

riddell-stan: Is there a reason for the `cd ..` in this script but not in the linux script?

riddell-stan:
On re-creating the linux multibuild experience, can we save the wheel as an artifact in some intermediate step and use it in a subsequent job?

.github/workflows/wheel_macOS.yml
ahartikainen: I should have removed that line

ahartikainen:
That should be possible. I think this works only inside one workflow file (this is fine for us).

https://help.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts#passing-data-between-jobs-in-a-workflow

So should we build wheel with multibuild and then install the wheel in another job (e.g ubuntu-18)


Btw, apparently one can run individual steps on specific containers or run all steps in a container (what we do).
https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idcontainer

riddell-stan:
We're very close with linux. I feel like macOS is done.

It would have been nice to have this linked to an issue where we could have clarified the goals. If the goal is to replace httpstan-wheels, we definitely need to go the last mile/kilometer and test the wheel in a non-manylinux2014 container. If the goal is something else, then I'm not sure how we would decide if it is finished.

riddell-stan:
Sorry! This wasn't supposed to be closed. It got automatically closed when I deleted the "master" branch with the new "main" branch.

@ahartikainen if you have trouble recovering or rebasing this, I can try to help.

PR #376, opened 2020-06-16
fix: Fix wheel building with empty extension module

riddell-stan:
The wheel building and repairing machinery recognizes packages as
containing platform-specific modules by the presence of distutils
Extension metadata. This commit adds an empty extension module so the
wheel building machinery is not confused. (httpstan has
platform-specific content (shared libraries) but no build-time extension
modules.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,9 @@ before_install:
   - if [[ -n $CC ]]; then $CC --version; fi
 install:
   - pip install "poetry<2,>=1.0" tox
+  # These are build requirements. poetry does not offer a way to specify build requirements.
   - pip install "mypy-protobuf==1.21"
+  - pip install "Cython<1,>=0.29"
   # this customization is just for `find` on macOS, can be removed after Stan 2.22 is used. See `Makefile` for details
   - |
     if [[ $TRAVIS_OS_NAME = "osx" ]]; then
--- a/build.py
+++ b/build.py
@@ -1,0 +1,10 @@
+from distutils.extension import Extension
+
+from Cython.Build import cythonize
+
+# empty extension module so build machinery recognizes package as platform-specific
+extensions = [Extension("httpstan.empty", sources=["httpstan/empty.pyx"])]
+
+
+def build(setup_kwargs):
+    setup_kwargs.update({"ext_modules": cythonize(extensions)})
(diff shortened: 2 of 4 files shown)


ahartikainen:
Is there any benefit with this compared to just use manual renaming (with suitable tool?)

riddell-stan:
I'm not sure if auditwheel or delocate will run at all if we do not do this.

riddell-stan:
But to answer your question more directly, we must do more than simply rename the file. For example, there's a file in the wheel called `WHEEL` and it contains a line:

``
Root-Is-Purelib: false
``

We would have to set this manually, I think. So we'd have to unzip the wheel, modify the file, and re-zip it.

I think this commit is a reasonable short-term fix. That said, I'd welcome a different solution that doesn't require such a hack.

.travis.yml
```diff
@@ -35,7 +35,9 @@ before_install:
   - if [[ -n $CC ]]; then $CC --version; fi
 install:
   - pip install "poetry<2,>=1.0" tox
+  # These are build requirements. poetry does not offer a way to specify build requirements.
   - pip install "mypy-protobuf==1.21"
+  - pip install "Cython<1,>=0.29"
```
ahartikainen, line 40: Is this defined here or inside poetry?

build.py
```diff
@@ -0,0 +1,10 @@
+from distutils.extension import Extension
+
+from Cython.Build import cythonize
+
+# empty extension module so build machinery recognizes package as platform-specific
+extensions = [Extension("httpstan.empty", sources=["httpstan/empty.pyx"])]
```
ahartikainen, line 6: Will there be a `httpstan.empty`?

Should we go with `_empty`

.travis.yml
riddell-stan: It's a build requirement. We need it to be installed before poetry calls build.py. There's no way to do this in poetry.

build.py
riddell-stan: It's not really a private module.

I think we're going to be stuck with this. It would be nice to translate empty.pyx into plain C so we didn't need to worry about installing Cython. One could also try to use pybind11.

ahartikainen: I think there is possibility to also create Extension without sources, but I think it might still need some compiled file, so calling it wouldn't crash.

PR #379, opened 2020-06-16
fix: Pin mypy-protobuf version, update mypy

### LINKED ISSUE #377: New version of mypy or mypy-protobuf causes test failure
This is the error which just emerged

```
httpstan/callbacks_writer_pb2.pyi:14: error: Module 'google.protobuf.internal.enum_type_wrapper' has no attribute '_EnumTypeWrapper'; maybe "EnumTypeWrapper"?
```

riddell-stan:
The newest version of mypy-protobuf 1.22 generates type annotations
which mypy does not recognize as valid.

Closes #377

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ before_install:
   - if [[ -n $CC ]]; then $CC --version; fi
 install:
   - pip install "poetry<2,>=1.0" tox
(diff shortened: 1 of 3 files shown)


riddell-stan:
Merged without review because this is trivial.

Issue #380, opened 2020-06-16
Flaky test failure related to stanc executable


riddell-stan:
There is a problem executing the `stanc` binary that shows up sometimes.

```
CRITICAL httpstan:views.py:99 Failed to compile module: [Errno 8] Exec format error: PosixPath('/home/travis/virtualenv/python3.7.6/lib/python3.7/site-packages/httpstan/stanc')
```

Superficially this looks like an error one would see if `stanc` were not marked as executable. But `stanc` is marked as executable -- otherwise all the test would fail on all platforms.

Could this be that the stanc binary is not being fully downloaded?

PR #381, opened 2020-06-16
fix: Do not download stanc silently

riddell-stan:
CI tests periodically fail due to some sort of problem with the `stanc`
executable. With this commit, stanc will be downloaded without
`-silent`.

Also added --retry and --retry-delay to mirror the approach used by
the stan repository.

Related to #380

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,14 +74,14 @@ $(HTTP_ARCHIVES_EXPANDED):
 ###############################################################################
 ifeq ($(shell uname -s),Darwin)
 build/stanc:
-	@curl --silent --location https://github.com/stan-dev/stanc3/releases/download/v$(STANC_VERSION)/mac-stanc -o $@
+	curl --location https://github.com/stan-dev/stanc3/releases/download/v$(STANC_VERSION)/mac-stanc -o $@ --retry 5 --fail
 else
 build/stanc:
-	@curl --silent --location https://github.com/stan-dev/stanc3/releases/download/v$(STANC_VERSION)/linux-stanc -o $@
+	curl --location https://github.com/stan-dev/stanc3/releases/download/v$(STANC_VERSION)/linux-stanc -o $@ --retry 5 --fail
 endif
 
 $(STANC): build/stanc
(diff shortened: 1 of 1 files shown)


ahartikainen:
Do we the hash for the file we download? I think we should compare against pre-calculated hash to determine the file is correct.

(This would needed to be done in stanc3 repo)

cc @rok-cesnovar

rok-cesnovar:
We could that, though not sure where that hash would reside. 

There was some discussion on this already for cmdstan, not sure its relevant: https://github.com/stan-dev/cmdstan/issues/755

riddell-stan:
I would hope that curl would fail if it didn't download completely.

This patch is intended to help discover what the problem is. If it is indeed curl doing something strange, perhaps there's some lightweight way to make sure it finishes. If there isn't, we could check against a hash.

rok-cesnovar:
Github occasionally (rare but it happens) returns error on requests for the stanc3 binary. I think it returns 403. We had this issue on cmdstan tests also: https://github.com/stan-dev/cmdstan/issues/837

riddell-stan:
I think ``--retry`` only applies to connection timeouts. That's not what we want.

Perhaps we need to use ``--retry-all-errors``?

riddell-stan:
Also, it looks like ``--retry-delay`` is not what we want. The default is a sensible exponential backoff. See https://stackoverflow.com/a/49825264

riddell-stan:
Since ``--retry-all-errors`` only exists in very recent versions of curl, we might not have a clean solution to this.

I think adding ``--fail`` might be a good idea in the meantime since at least we would understand what happened (when GitHub responded with a 403 or 404 or something).

rok-cesnovar:
I think if you apply what Cmdstan has, you will be fine. We had 0 tests faile due to failure to download stanc3 and we run a bunch of tests daily. Previously, we had issues almost every day.

ahartikainen:
You might want to add `--connect-timeout` to retry with timeout case, but I don't remember what error github gives when download fails.

riddell-stan:
I added ``--fail``. This is probably better than nothing.

PR #384, opened 2020-06-19
test: Simplify test fixtures

riddell-stan:
Big simplification of test setup thanks to pytest-asyncio.
No need for running the server in a separate thread.

(--forked may still be needed for macOS, we will see. I have no way of testing this right now)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,17 +10,13 @@ jobs:
   - python: 3.7
     env:
       - CC=clang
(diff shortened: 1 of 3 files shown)


riddell-stan:
"+6 −25" if only every PR looked like this.

riddell-stan:

Contributors should feel free to revert any commits in this pull request for any reason or for no reason at all. You can revert a commit with the git revert command.

PR #386, opened 2020-06-20
Refactor model-specific stan::services extension module building

### LINKED ISSUE #385: Creating a temporary directory from which to import the Stan model extension module is wasteful
Not only is it wasteful, it seems to be leading to a race condition in the multiprocessing setup (observed in PyStan 3 tests recently).

I believe it is possible to compile a Stan model extension module (the kind backed by `anonymous_stan_model_services.pyx.template`) once, save it to the `~/.cache/httpstan/2.n.n/models/{ model_id }` folder and import it from there every time it is used.

If this works *many* lines of code could be removed. The software would be considerably easier to maintain. We may even be able to avoid using sqlite3 and just use files on disk.

riddell-stan:
This is a big refactor.

Goal: Compile a model-specific stan::services-wrapping extension module (the kind backed by anonymous_stan_model_services.pyx.template) once, saving it in ~/.cache/httpstan/n.n.n/models/{ model_id } folder. It will be imported from this location every time it is used.

This eliminates the need to save the extension module in a database and extract it into a temporary directory every time it is used.

Closes #385

### DIFF
--- a/httpstan/build_ext.py
+++ b/httpstan/build_ext.py
@@ -1,0 +1,107 @@
+"""Lightly modified build_ext which captures stdout and stderr.
+
+isort:skip_file
+"""
+
+# IMPORTANT: `import setuptools` MUST come before any module imports `distutils`
+# background: https://bugs.python.org/issue23102
+import setuptools  # noqa: F401
+
+import distutils.command.build_ext
+import distutils.core
+import io
+import os
+import sys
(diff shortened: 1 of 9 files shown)


riddell-stan:

Contributors should feel free to revert any commits in this pull request for any reason or for no reason at all. You can revert a commit with the git revert command.

PR #387, opened 2020-06-23
chore: Bump version

riddell-stan:


riddell-stan:
This new version appears to be working with my local copy of PyStan 3.

riddell-stan:
Merging without review because it's a trivial change and tests pass.

PR #388, opened 2020-06-24
fix: Do not download stanc silently

### LINKED ISSUE #380: Flaky test failure related to stanc executable

riddell-stan:
CI tests periodically fail due to some sort of problem with the `stanc`
executable. With this commit, stanc will be downloaded without
`-silent`.

Closes #380

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,14 +74,14 @@ $(HTTP_ARCHIVES_EXPANDED):
 ###############################################################################
 ifeq ($(shell uname -s),Darwin)
 build/stanc:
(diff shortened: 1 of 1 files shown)


riddell-stan:
This was approved earlier. (Re-created due to main branch truncation/renaming.)

PR #389, opened 2020-06-24
Upload wheels to pypi

ahartikainen:
This PR builds httpstan wheels with Github Actions when any tag is created.

- [x] wheel: manylinux2014
- [x] wheel: macOS
- [x] new secrets: `PYPI_PASSWORD` and `TEST_PYPI_PASSWORD` (done in github)

### DIFF
--- a/.github/workflows/wheel_macOS.yml
+++ b/.github/workflows/wheel_macOS.yml
@@ -1,0 +1,69 @@
+name: wheel - macOS
+
+on:
+  push:
+    tags:
+      - '*'
+jobs:
+  wheel_macos:
+    name: macOS wheel
+    runs-on: ${{ matrix.os }}
+    strategy:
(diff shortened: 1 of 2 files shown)


ahartikainen:
Original PR https://github.com/stan-dev/httpstan/pull/374

ahartikainen:
There is currently a small problem with macOS deployment target. It should be 1.9 but .configure(?) runs with 1.14.

I think that problems comes from the python version which is built against 1.14.

I will see if there is an alternative way to get another python source (I think python.org should work. ->> I was under the impression that get python action was using python.org versions)

ahartikainen:
I close this for now.

The wrong python on osx (non-framework) is not good. And fixing it is not that simple as it should be. (I think we need to build against 37 and 38).

PR #393, opened 2020-07-05
docs: Align docs with PyStan

riddell-stan:
One significant change here. "How to contribute" is moved into pystan.
The code contribution standards are the same for httpstan and pystan.


doc/installation.rst
```diff
@@ -4,8 +4,35 @@ Installation
 
 .. These instructions appear in both README.rst and installation.rst
 
-**httpstan** requires Python ≥ 3.6. macOS and Linux are supported.
+**httpstan** requires Python ≥ 3.7. macOS and Linux are supported.
 
 ::
 
     $ python3 -m pip install httpstan
```
ahartikainen, line 11: is this python3 use some Linux system specific use? I think ubuntu for example defaults to python 3 with `python`.

riddell-stan: `python` does not exist on newer versions of ubuntu. It's just python3.

PR #397, opened 2020-07-12
ci: Remove in-tree shared libraries before tests

riddell-stan:
Remove in-tree shared libraries before running tests. httpstan
must use shared libraries which are contained in the httpstan wheel.
By removing all other copies of the shared libraries, we make sure this
happens.


riddell-stan:
This is unnecessary and it adds lines of code to read.

PR #398, opened 2020-07-12
fix: Remove hard-coded LC_ID_DYLIB on macOS

riddell-stan:
protobuf macOS shared libraries have a LC_ID_DYLIB path which is
fixed at compile time. Since we move the protobuf shared library after
compiling it, this creates problems. This commit fixes the problem.

This should fix https://github.com/stan-dev/pystan-next/issues/128


riddell-stan:
macOS probably shouldn't be supported since we don't have at least two developers who are willing and able to test macOS-specific Python code.

It certainly would be nice to have someone else working on this.

riddell-stan:

PR #400, opened 2020-07-14
fix: Remove unused Cython code

riddell-stan:
stanc3 is used now, compile.pyx concerns stanc2.


riddell-stan:

PR #401, opened 2020-07-14
feat: Remove Cython build dependency

### LINKED ISSUE #399: Cython is a (package) build-time dependency
When we are building the package we use Cython for `empty.pyx`. In the short term, we cannot get rid of this placeholder extension module. But we can write it using pybind11 or pure C, eliminating the build-time dependency on Cython. Build dependencies are hard to specify using poetry and somewhat confusing in general.

riddell-stan:
Remove inconvenient build-time dependency on Cython.
Cython is used by httpstan to wrap the model-specific stan::services
functions. Cython was also used to build a module "empty" whose
only purpose was to force the Python wheel building tools to
recognize the httpstan wheel as platform-specific (which it is due
to the presence of shared libraries in `httpstan/lib`).

Closes #399


riddell-stan:

PR #405, opened 2020-07-14
chore: Update version to 2.0.3

riddell-stan:
Update version to 2.0.3. macOS wheels should be fixed.


riddell-stan:
Merged without review. This is a trivial change.

PR #406, opened 2020-07-14
fix: Version defined in pyproject.toml only

riddell-stan:
Uses the Python 3.8 function `importlib.metadata.version` to
retrieve version information from the installed package.

Aligns httpstan code with pystan.


riddell-stan:
I think we're fine for now. Python 3.7 users simply can't access ``httpstan.__version__``, which doesn't seem like a particularly big problem.

PR #410, opened 2020-07-20
fix: Store current operations in memory

riddell-stan:
Store current operations in memory, not in sqlite. There is
no need to persist operations. Using sqlite is virtually
certain to be slower. Using sqlite also adds complexity:
creating the table, opening and closing a connection.

We do not use threads so there is no thread-safety
concern.

The load_fit and dump_fit functions are now not async. They never needed to be async.


riddell-stan:

PR #411, opened 2020-07-20
refactor: Write fit messages in single place

### LINKED ISSUE #352: Create fit handler is needlessly complex
`handle_create_fit` in views.py is 154 lines of code and involves two different callbacks. It should be simplified.

- [x] ~Eliminate the poorly-written callback `_services_call_done`. The operation id can be sent back to the client before the handler itself returns. See <https://docs.aiohttp.org/en/latest/faq.html#how-do-i-perform-operations-in-a-request-handler-after-sending-the-response>.~ (Not possible, can't really close the request doing this.)
- [ ] Have the `call` function in `services_stub.py` write the logger and writer messages to disk. Having two functions (`call` and `handle_create_fit`) both involved in writing these messages to disk is more confusing than having a single function perform the work.

riddell-stan:
The services_stub.call function takes responsibility for
writing the messages it gets from Stan to disk. Previously
the `call` function did some writing but closing the buffer
was handled elsewhere.

Closes #352

Assumes #410 is merged.

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -9,6 +9,7 @@ Unix domain socket.
 import asyncio
 import concurrent.futures
 import functools
+import io
 import logging
 import os
 import select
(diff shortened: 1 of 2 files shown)


riddell-stan:

PR #412, opened 2020-07-22
fix: Do not starve asyncio loop during sampling

riddell-stan:
Avoid effectively blocking the asyncio loop during sampling by
`sleep`ing briefly when there are no messages to process. Previously
`select` would completely monopolize the CPU until sampling had
completed. As a result, the server would not process simple GET
requests, including the request to learn the current sampling progress.
With this change, such requests are processed.

Second commit contains a very minor performance improvement.

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -116,11 +116,14 @@ async def call(
                 messages_file.write(message)
             # if `potential_readers == [socket_]` then either (1) no connections
             # have been opened or (2) all connections have been closed.
-            if not readable and potential_readers == [socket_]:
-                if future.done():  # type: ignore
-                    logger.debug(f"Stan services function `{function_basename}` returned or raised a C++ exception.")
+            if not readable:
+                if potential_readers == [socket_] and future.done():  # type: ignore
+                    logger.debug(
+                        f"Stan services function `{function_basename}` returned without problems or raised a C++ exception."
+                    )
                     break
-                await asyncio.sleep(0.01)
+                # no messages right now and not done. Sleep briefly so other pending tasks get a chance to run.
+                await asyncio.sleep(0.001)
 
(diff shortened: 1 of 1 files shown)


httpstan/services_stub.py
```diff
@@ -116,11 +116,14 @@ async def call(
                 messages_file.write(message)
             # if `potential_readers == [socket_]` then either (1) no connections
             # have been opened or (2) all connections have been closed.
-            if not readable and potential_readers == [socket_]:
-                if future.done():  # type: ignore
-                    logger.debug(f"Stan services function `{function_basename}` returned or raised a C++ exception.")
+            if not readable:
+                if potential_readers == [socket_] and future.done():
+                    logger.debug(
+                        f"Stan services function `{function_basename}` returned without problems or raised a C++ exception."
```
ahartikainen, line 122: What other options there are?

- no problems
- C++ exception

riddell-stan: Those are the only options, I think.

riddell-stan:
The 0.01 second wait in `select` doesn't use the CPU. But during this time Python cannot do anything. If we set it to 0 this means, in theory, `httpstan` could be processing other requests (instead of waiting for messages from Stan).

We probably should not worry about this too much until we have a more systematic approach to benchmarking things. The first commit (moving the `asyncio.sleep` around) is really important though. The progress bar in pystan does not work without it.

httpstan/services_stub.py
riddell-stan: There might be a third option: the future could be canceled. This might happen if the subprocess is killed by some external program. Maybe an out of memory error would also kill the process and cancel the future.

If there's an easy way to artificially (sharply) limit the amount of memory available to a process, we could test this.

PR #419, opened 2020-08-03
chore: Bump version

riddell-stan:


riddell-stan:
Merged without review as this is a trivial change.

PR #421, opened 2020-08-07
refactor: Align Cython code better with CmdStan

riddell-stan:
Make minor changes to better align Cython code with Cmdstan's
command.hpp. If and when people are interested in understanding how this
wrapping code works it will be helpful to have all code which
does the same thing look approximately the same. For example,
CmdStan references the model with an instance of model_base, so now we
do as well. No changes here change behavior.


riddell-stan:
This is a really simple refactor. I'm bothering with it because I think it will make rewriting things using pybind11 a tiny bit easier.

PR #423, opened 2020-08-10
fix: Remove unused code

riddell-stan:
Remove old Cython code which is no longer used.

### DIFF
--- a/httpstan/compile.py
+++ b/httpstan/compile.py
@@ -1,7 +1,9 @@
 """Runs the `stanc` binary in a subprocess to compile a Stan program."""
 import importlib.resources
+import os
 import subprocess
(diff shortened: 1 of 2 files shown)


ahartikainen:
Do we need to add definition for 

    with importlib.resources.path(__package__, "stanc") as stanc_binary:

as 

    stanc_binary: pathlib.Path
    with importlib.resources.path(__package__, "stanc") as stanc_binary:

riddell-stan:
That's something to consider. It's definitely a different issue though.

riddell-stan:
This seems like a mypy bug. stanc_binary is interpreted as a Path.

riddell-stan:

PR #424, opened 2020-08-10
feat: Return stanc warning messages

### LINKED ISSUE #273: Logger info, warn message tests missing
A test is needed to check that non-fatal info and warning messages are sent to the client.

The test will likely be similar to the one used in the fatal case #272 .

Once this test is added, a test on the pystan next side will be needed.

Here's a simple example of something the user needs to see:

```
Info: assignment operator <- deprecated in the Stan language; use = instead.
```

Edit: add example

### LINKED ISSUE #349: Pedantic mode warnings not returned to user
(This assumes stanc3 is being used.)

The pedantic mode warnings are useful. They need to be accessible.

riddell-stan:
Return stanc warning messages when stanc compiles Stan program code
successfully. Previously program code which compiled successfully
with warnings would raise an exception.

Closes #349
Closes #273


httpstan/views.py
```diff
@@ -85,6 +85,15 @@ async def handle_models(request: aiohttp.web.Request) -> aiohttp.web.Response:
 
     program_code = args["program_code"]
     model_name = httpstan.models.calculate_model_name(program_code)
+
+    # compile `program_code` twice, first time only to check for errors and get warnings
+    try:
+        stanc_warnings = await httpstan.models.stanc_warnings(program_code)
+    except ValueError as exc:
+        message, status = f"Exception while compiling `program_code`: `{repr(exc)}`", 400
```
ahartikainen, line 93: This looks good.

Returning traceback here might be a bit overkill.

ahartikainen:
We need to merge #423 before this, so typing will pass.

PR #426, opened 2020-08-10
feat: Add debug mode

### LINKED ISSUE #390: Debugging segfaults is difficult
Debugging segfaults is really difficult when we run the sampler in a different process. By contrast, if everything runs in a process with no multithreading/multiprocessing, debugging with `gdb` is remarkably easy.

Here's a solution. If the user sets the environment variable ``HTTPSTAN_DEBUG`` to a truthy value, disable all multiprocessing/multithreading associated to the call to the stan::services functions. This means that the call to the ``fits`` endpoint will block. No parallel sampling will be possible.

If ``HTTPSTAN_DEBUG`` is set, one would also *not* redirect stderr or stdout.

riddell-stan:
Add a debug mode which can be enabled via the HTTPSTAN_DEBUG environment variable.
If `HTTPSTAN_DEBUG` is set, httpstan will not run sampling calls in a
background thread or process. This makes debugging via `gdb` much
easier.

Closes #390

### DIFF
--- a/doc/developer_resources.rst
+++ b/doc/developer_resources.rst
@@ -11,6 +11,18 @@ If CmdStan changes default values for arguments, pipe the output of ``modelbinar
 file and then point ``scripts/parse_cmdstan_help.py`` at that file. The output
 should replace the file ``httpstan/services/cmdstan-help-all.json``.
 
+DEBUG mode
+==========
+
+If the environment variable ``HTTPSTAN_DEBUG`` is set to ``1`` or ``true``, the
+call to the stan::services sampling function will block until finished instead
+of being run in the background. This should make debugging crashes with ``gdb``
+much easier.  There is, however, one complication. The number of samples drawn
+must be set to be a very low number (e.g., ``10``) otherwise the call to the
+sampling function will freeze up. The call will freeze up because the socket
+used to communicate from C++ to Python will fill up. If the socket runs out of
+buffer space the stan::services call will never return.
+
 Signing key
 ===========
 The signing key for httpstan has id ``CB808C34B3BFFD03EFD2751597A78E5BFA431C9A``. Git tags are signed with this key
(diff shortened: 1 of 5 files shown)


doc/developer_resources.rst
```diff
@@ -11,6 +11,18 @@ If CmdStan changes default values for arguments, pipe the output of ``modelbinar
...
 should replace the file ``httpstan/services/cmdstan-help-all.json``.
 
+DEBUG mode
+==========
+
+If the environment variable ``HTTPSTAN_DEBUG`` is set to ``1`` or ``true``, the
+call to the stan::services sampling function will block until finished instead
+of being run in the background. This should make debugging crashes with ``gdb``
+much easier.  There is, however, one complication. The number of samples drawn
+must be set to be a very low number (e.g., ``10``) otherwise the call to the
```
ahartikainen, line 21: This is fine

Is there a way to manually increase the socket size or automatically empty it?

httpstan/services_stub.py
```diff
@@ -89,7 +90,13 @@ async def call(
 
         lazy_function_wrapper = _make_lazy_function_wrapper(function_basename, model_name)
         lazy_function_wrapper_partial = functools.partial(lazy_function_wrapper, socket_filename.encode(), **kwargs)
-        future = asyncio.get_running_loop().run_in_executor(executor, lazy_function_wrapper_partial)
+
+        # If HTTPSTAN_DEBUG is set block until sampling is complete. Do not use an executor.
+        if HTTPSTAN_DEBUG:
+            future: asyncio.Future = asyncio.Future()
+            future.set_result(lazy_function_wrapper_partial())
```
ahartikainen, line 97: This looks good

riddell-stan:
That's a great idea. I'll make the addition.

doc/developer_resources.rst
riddell-stan: I did give this some thought. I don't think there's an easy way.

We can print a warning about the requirement that num_samples be set to 10 though (when HTTPSTAN_DEBUG is set). That should help a great deal.

riddell-stan:
I rebased and added some warning messages.

PR #433, opened 2020-08-27
ci: Remove codecov.yml

ahartikainen:
This PR removes codecov.yml file.

This means that codecov won't send a comment to a PR for the current linehit situation, but developer can still go to https://codecov.io/gh/stan-dev/httpstan and see the situation if needed.


riddell-stan:
Thanks.

Any chance we could remove the other lines in the config file -- or add 
comments explaining why they are there? (or explain in this issue?)

ahartikainen:
Yes, it is true that we don't need the file. I personally use the site to see which parts of the code are used.

PR #440, opened 2020-09-01
ci: Remove unused coverage config, docs

riddell-stan:
Remove unused codecov line in travis config.
Remove reference to unused coveralls service in docs.


ahartikainen:
We don't want to upload covarage to codecov?

riddell-stan:
We have a floor on coverage, we require at least 93% coverage with the `--cov-fail-under=93` argument to pytest. So new PRs which don't have tests will, in theory, fail.

Until someone adopts the task of bringing coverage up to 100%, I think it makes sense to remove this. We already use pytest-cov. I'm not sure what additional coverage software adds.

Also, running a unverified bash script one downloads off the internet is not a recommended security practice. There was a discussion on the Stan discourse forum about something similar recently.

ahartikainen (approved):

PR #444, opened 2020-09-04
feat: compress/decompress cache with gzip

ahartikainen:
Use `gzip` to compress and decompress cache with the default settings.

### DIFF
--- a/httpstan/cache.py
+++ b/httpstan/cache.py
@@ -2,6 +2,7 @@
 
 Functions in this module manage the Stan model cache and related caches.
 """
+import gzip
 import logging
 import os
 import pathlib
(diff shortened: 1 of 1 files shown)


httpstan/cache.py
```diff
@@ -47,7 +50,7 @@ def dump_fit(name: str, fit_bytes: bytes) -> None:
     fit_path = os.path.join(*([cache_path] + name.split("/")[:-1]))
     fit_filename = os.path.join(fit_path, f'{name.split("/")[-1]}.dat')
```
riddell-stan, line 51: I'd add .gz to the suffix.

httpstan/cache.py
```diff
@@ -69,7 +74,7 @@ def load_fit(name: str) -> bytes:
...
```
riddell-stan, line 75: same comment as above.

httpstan/cache.py
ahartikainen: thanks, yes we should do that

riddell-stan (approved):

PR #450, opened 2020-09-09
fix: Update webargs to version 6

riddell-stan:
This newer version of webargs resolves a long-standing validation
issue.


tests/test_sampling_exceptions.py
```diff
@@ -29,7 +29,7 @@ async def test_sampling_initialization_failed(api_url: str) -> None:
     operation = await helpers.sample(api_url, program_code, payload)
 
     # verify an error occurred
-    assert operation["result"].get("code") == 400
+    assert operation["result"].get("code") == 400, operation["result"]
```
ahartikainen, line 32: will this work?

Is this same as 

    assert bool, bool

    assert True, False --> assertion is not raised

ahartikainen (commented):
Added one question for `assert`

riddell-stan:
Good catch. That shouldn't be there.

ahartikainen (approved):

PR #457, opened 2020-09-16
feat: Add list_models endpoint

### LINKED ISSUE #409: No way to check if model is in cache (via HTTP)
There should be a way for a client to check to see if a model has already been compiled. There currently is no way to do this. See https://github.com/stan-dev/pystan-next/pull/132 for a workaround.

Here's one approach that could fit with the Google REST API style:
- Create a calculate-model-name endpoint (check to see if there's a recommended naming pattern for this sort of thing in the google API style guide) which exposes `httpstan.models.calculate_model_name`.
- Create an endpoint that lists all models which are in the cache.

riddell-stan:
Allows users to list models in cache. REST APIs
typically offer such a list.

Such a list is required if a user wants to delete
all models from the cache. This is because, without
this function, there is no way for the user to know
what models are in the cache. Models must be referred
to by name in order to delete them.

Closes #409

### DIFF
--- a/httpstan/cache.py
+++ b/httpstan/cache.py
@@ -7,6 +7,8 @@ import logging
 import os
 import pathlib
 import shutil
+import typing
+from importlib.machinery import EXTENSION_SUFFIXES
 
 import appdirs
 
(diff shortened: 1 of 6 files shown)


ahartikainen (approved):

riddell-stan:
Tests are passing. Not sure why there are two "Travis CI - Pull request" lines above in GitHub.

PR #459, opened 2020-09-19
feat: Allow users to delete a model

riddell-stan:
Allow users to delete a model. Deleting a model
also removes all associated fits.


riddell-stan:

PR #465, opened 2020-10-03
feat: Remove Cython code, use pybind11 instead

### LINKED ISSUE #403: Stan C++ wrapping code is written in Cython, not pybind11
All Cython code should be removed and replaced with C++ code using pybind11. This would eliminate Cython as one of the languages required to understand code in this repository.

The code which needs to be replaced is in the file `httpstan/anonymous_stan_model_services.pyx.template`.

riddell-stan:
Remove all Cython code, replacing the code with
C++ (using pybind11).

One feature which Cython supported that pybind11 does not is function
signature inspection.  Whereas you can call
`inspect.Signature.from_callable` on a Cython-defined function, you
cannot do this with pybind11-defined functions.
So we fall back to parsing the docstring.

This commit also formats C++ files with:

clang-format -i -style="{ColumnLimit: 119}"

Closes #403 

Removes 129 lines of code. 🚀


httpstan/models.py
```diff
@@ -84,28 +83,17 @@ def import_services_extension_module(model_name: str) -> ModuleType:
         module_path = next(filter(lambda p: p.suffix in EXTENSION_SUFFIXES, model_directory.iterdir()))
     except (FileNotFoundError, StopIteration):
         raise KeyError(f"No module for `{model_name}` found in `{model_directory}`")
-    module_name = pathlib.Path(module_path.stem).stem
-    spec = importlib.util.spec_from_file_location(module_name, module_path)  # type: ignore
+    # The module name, which is independent of the filename, is always "stan_services". The module
+    # name must be defined in stan_services.cpp, which is compiled before we know with which
+    # specific stan model it will be linked with. Since we want to compile stan_services.cpp in
+    # advance, we we are stuck with a fixed module name.
```
ahartikainen, line 89: we we

ahartikainen (approved):
LGTM

Found only 1 typo

httpstan/models.py
riddell-stan: Thanks. Fixed.

riddell-stan:
Tests are passing. CI failure due to failure to install/download isort.

PR #466, opened 2020-10-03
ci: Hotfix, avoid poetry 1.1.0

riddell-stan:
Bug which affects `poetry export` is unfixed in 1.1.0.
Fixed in future releases.
See https://github.com/python-poetry/poetry/issues/3063


riddell-stan:
Trivial fix. Merged without review.

PR #467, opened 2020-10-04
Precompile stan_services.cpp

riddell-stan:
This will dramatically speed up compilation of the model-specific Python extensions. The strategy here mirrors the strategy used by CmdStan.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -42,7 +42,7 @@ LIBRARIES += httpstan/lib/libprotobuf-lite.so
 endif
 INCLUDES_STAN_MATH_LIBS := httpstan/include/lib/boost_$(BOOST_VERSION) httpstan/include/lib/eigen_$(EIGEN_VERSION) httpstan/include/lib/sundials_$(SUNDIALS_VERSION) httpstan/include/lib/tbb_$(TBB_VERSION)
 INCLUDES_STAN := httpstan/include/stan httpstan/include/stan/math $(INCLUDES_STAN_MATH_LIBS)
-INCLUDES := httpstan/include/google/protobuf httpstan/include/pybind11 $(INCLUDES_STAN)
+INCLUDES := httpstan/include/google httpstan/include/pybind11 $(INCLUDES_STAN)
 STANC := httpstan/stanc
 PRECOMPILED_OBJECTS = httpstan/callbacks_writer.pb.o httpstan/stan_services.o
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -42,11 +42,11 @@ LIBRARIES += httpstan/lib/libprotobuf-lite.so
 endif
 INCLUDES_STAN_MATH_LIBS := httpstan/include/lib/boost_$(BOOST_VERSION) httpstan/include/lib/eigen_$(EIGEN_VERSION) httpstan/include/lib/sundials_$(SUNDIALS_VERSION) httpstan/include/lib/tbb_$(TBB_VERSION)
 INCLUDES_STAN := httpstan/include/stan httpstan/include/stan/math $(INCLUDES_STAN_MATH_LIBS)
-INCLUDES := httpstan/include/google/protobuf httpstan/include/pybind11 $(INCLUDES_STAN)
+INCLUDES := httpstan/include/google httpstan/include/pybind11 $(INCLUDES_STAN)
```
ahartikainen, line 45: Do we have more than 1 lib in google folder?

ahartikainen (approved):
LGTM

I added some notes.

Makefile
riddell-stan: `httpstan/include/google` is consistent with `httpstan/include/pybind`, the source code is included with <google/protobuf/...>.

Another problem here is that "INCLUDES" is not a good variable name since it does not give us anything equivalent to `include_dirs` in distutils terms.

Makefile
```diff
@@ -227,3 +227,35 @@ endif
...
+###############################################################################
+# Precompile httpstan-related objects, eventually linked in httpstan/models.py
+###############################################################################
+
+
+PYTHON_CC ?= $(shell python3 -c 'import sysconfig;print(" ".join(sysconfig.get_config_vars("CC")))')
+PYTHON_CFLAGS ?= $(shell python3 -c 'import sysconfig;print(" ".join(sysconfig.get_config_vars("CFLAGS")))')
+PYTHON_CCSHARED ?= $(shell python3 -c 'import sysconfig;print(" ".join(sysconfig.get_config_vars("CCSHARED")))')
+PYTHON_INCLUDE ?= -I$(shell python3 -c'import sysconfig;print(sysconfig.get_path("include"))')
+PYTHON_PLATINCLUDE ?= -I$(shell python3 -c'import sysconfig;print(sysconfig.get_path("platinclude"))')
```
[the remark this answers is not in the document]
riddell-stan: I don't know what the best practice here is. I thought it was to always use `python3`. This might be a Linux thing since the current version of Debian/Ubuntu does not even provide a `python` executable.

PR #468, opened 2020-10-04
ci: Verify that httpstan does not break pystan

riddell-stan:
Verify that httpstan does not break pystan by
importing pystan and sampling from a simple model.


riddell-stan:

PR #475, opened 2020-10-06
ci: Use working version of poetry

riddell-stan:
Newer versions of poetry do not export dev-dependencies extras
correctly. See https://github.com/python-poetry/poetry/issues/2909


riddell-stan:
Trivial fix. Merged without review.

PR #476, opened 2020-10-06
chore: Use newer versions of packages

riddell-stan:
Use newer versions of the following packages:

- apispec
- numpy
- protobuf

Choice of numpy 1.16 reflects:

https://numpy.org/neps/nep-0029-deprecation_policy.html


ahartikainen (approved):

riddell-stan:
Error is due to Python 3.9 not being available on macOS yet.

PR #477, opened 2020-10-07
fix: Use g++ instead of gcc when pre-compiling

riddell-stan:
Use g++ instead of gcc when pre-commpiling C++ code.
This does not appear to matter with gcc 9.3 but may
matter with older versions of gcc.


riddell-stan:
Minor fix. Merged without review.

PR #489, opened 2020-11-01
test: Verify (de)serialization of nan, inf, -inf

riddell-stan:
Verify that nan, inf, and -inf values are correctly
serialized.

### DIFF
--- a/tests/test_nan_inf.py
+++ b/tests/test_nan_inf.py
@@ -1,0 +1,38 @@
+"""Test serialization of nan and inf values."""
+import math
+
+import pytest
+
+import helpers
+
+program_code = """
+    parameters {
+      real eta;
+      real positive_infinity;
+      real negative_infinity;
+      real not_a_number;
+    }
+    transformed parameters {
+      real alpha;
+      real beta;
+      real gamma;
+      alpha = not_a_number();
+      beta = positive_infinity();
+      gamma = negative_infinity();
+    }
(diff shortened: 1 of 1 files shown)


tests/test_nan_inf.py
```diff
@@ -0,0 +1,38 @@
...
+import math
+
+import pytest
+
+import helpers
+
+program_code = """
+    parameters {
+      real eta;
+      real positive_infinity;
```
ahartikainen, line 11: why are these defined in parameters block? Aren't these special functions?

riddell-stan: I don't know. It seems like this is how it is done in the Stan language. See https://github.com/stan-dev/stan/blob/develop/src/test/test-models/good/nullary-unconflicted.stan

If there's a better way, I'm happy to change things. This seems to work.

ahartikainen: I was just reading from here https://mc-stan.org/docs/2_25/functions-reference/special-values.html

ahartikainen (approved):

tests/test_nan_inf.py
riddell-stan: I'm really short on time and I'd like to get httpstan and pystan fixed. If you think there's a better way to write this test---and there certainly may be a better way---perhaps we could discuss it in a new issue?

PR #490, opened 2020-11-01
test: Verify a model with many parameters works

riddell-stan:
Check that a model with many parameters (160) works.
Prompted by a unusual bug in late 2020.

### DIFF
--- a/tests/test_large_model.py
+++ b/tests/test_large_model.py
@@ -1,0 +1,40 @@
+"""Test sampling from a model with many parameters."""
+import pytest
+
+import helpers
+
+program_code = """
+    data {
+      int<lower=0> J; // number of schools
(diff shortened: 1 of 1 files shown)


riddell-stan:
- [ ] This test actually needs to be in pystan because the decoding issue was there. (Adding it here does no harm.)

ahartikainen:
This test should make it sure that in fit_bytes has correct amount of parameters and they are in correct order. This is what failed in pystan, but is already in fit_bytes.

riddell-stan:
Looks like the problem is with httpstan and the unix socket buffer size (on macOS). This test should be useful.

riddell-stan:
This reproduces the error in https://github.com/stan-dev/pystan-next/issues/163. Test is appropriately failing.

PR #491, opened 2020-11-03
BREAKING CHANGE: Use bytes not strings in protobuf messages

riddell-stan:
Use `bytes` not `string`s in Protocol Buffers messages. According to the
protocol buffers specification, `string`s may only be used when one can
guarantee that the underlying bytes are valid UTF-8.  (We use C++
`std::string`s and perform no checking that anything is valid UTF-8.)
According to the spec, we must use Protocol Buffer `bytes`.

This may address a mysterious bug on macOS that seems to involve
a failure in the encoding (or decoding) of a protocol buffer message.

This is a breaking change since we are changing the protocol buffer
schema.

Note for reviewers: There are no substantive changes here except replacing protobuf `string` types with protobuf `bytes` types.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -65,7 +65,7 @@ def extract(param_name: str, fit_bytes: bytes) -> typing.List[typing.Union[int, 
         assert msg
         if msg.topic == callbacks_writer_pb2.WriterMessage.Topic.Value("SAMPLE"):
             for value_wrapped in msg.feature:
-                if param_name == value_wrapped.name:
+                if param_name == value_wrapped.name.decode():
                     fea = getattr(value_wrapped, "double_list") or getattr(value_wrapped, "int_list")
                     draws.append(fea.value.pop())
     if len(draws) == 0:
--- a/README.rst
+++ b/README.rst
@@ -118,7 +118,7 @@ decoded message should resemble the following::
 
     topic: LOGGER
     feature {
-      string_list {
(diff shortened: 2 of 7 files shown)


riddell-stan:
This is ready to be merged. The failing test is due to the pystan test failing (which it should, since pystan needs to be updated to use the new protobuf message definition). The semantic pull request checker is failing because it doesn't recognize "BREAKING CHANGE:" as a prefix, but should.

tests/helpers.py
```diff
@@ -65,7 +65,7 @@ def extract(param_name: str, fit_bytes: bytes) -> typing.List[typing.Union[int,
         assert msg
         if msg.topic == callbacks_writer_pb2.WriterMessage.Topic.Value("SAMPLE"):
             for value_wrapped in msg.feature:
-                if param_name == value_wrapped.name:
+                if param_name == value_wrapped.name.decode():
```
ahartikainen, line 68: This is utf-8 by default?

ahartikainen (approved):
Good change regardless of any bugs.

tests/helpers.py
riddell-stan: Yes, I believe utf-8 has been the default for some time. https://docs.python.org/3/library/stdtypes.html?highlight=decode#bytes.decode Is that what you mean?

ahartikainen: Yes

PR #493, opened 2020-11-17
BREAKING CHANGE: Use JSON instead of Protocol Buffers

riddell-stan:
Use JSON instead of Protocol Buffers for serializing messages from Stan.
Simplifies code considerably.

TODOs:
- [x] Update docs


riddell-stan:
I had been planning on making this change for some time. There's not much a performance cost (if we use simdjson on the PyStan side). (In fact, simdjson seems to read faster than protobuf in most tests in Python.) The code is much easier to maintain since nobody needs to know about protocol buffers.

This PR needs no real justification, however, since we have a bug in protobuf with no known fix, https://github.com/stan-dev/pystan-next/issues/163 . Not having pystan work on macos is not really an option.

ahartikainen (approved):
LGTM

Just a minor question, how rapidjson handles nan / -inf / inf?

riddell-stan:
It uses the same strategy as Python. You do have to set a special flag, however. Here's a line where we use the flag (``kWriteNanAndInfFlag``).

```
rapidjson::Writer<rapidjson::StringBuffer, rapidjson::UTF8<>, rapidjson::UTF8<>, rapidjson::CrtAllocator, rapidjson::kWriteNanAndInfFlag> writer(buffer);
```

PR #498, opened 2020-11-26
ci: github actions

ahartikainen:
Run tests in GitHub Actions, testing against Linux (Ubuntu 20.04) and macOS (10.15 and 11.0).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,51 @@
+name: httpstan
+
+on: [push, pull_request]
+
+jobs:
+  tests:
+    name: httpstan tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.14]
+        python-version: [3.7, 3.8]
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install poetry
+        run: pip install "poetry==1.0.9"
+
+      - name: Build libraries used by Stan
+        run: make
+
+      # export test dependencies from pyproject.toml, install them
+      - name: Install dependencies
+        run: |
+          poetry export -f requirements.txt --dev -o requirements.txt \
+          && pip install -r requirements.txt
+
+      - name: Check code
+        run: scripts/check
+
+      - name: Build and Install wheel
+        run: |
+          poetry build -v
+          python -m pip install dist/*.whl
+
+      - name: Run tests
+        run: |
+          mkdir testdir
+          cd testdir
+          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 ../tests
+
+      - name: Test PyStan compatibility
+        run: |
+          pip install --pre pystan
+          python -c'import stan;assert stan.build("parameters {real y;} model {y ~ normal(0,1);}").sample()["y"] is not None'


ahartikainen:
For Action results see https://github.com/ahartikainen/httpstan/actions

These are not currently shown here until the main repo has `.github/workflows/main.yml` file.

ahartikainen:
Runner versions can be found here

https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners

ahartikainen:
Ok, tests are passed.

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+          poetry build -v
+          python -m pip install dist/*.whl
+
+      - name: Run tests
+        run: |
+          mkdir testdir
+          cd testdir
+          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 ../tests
+
+      - name: Test PyStan compatibility
```
riddell-stan, line 59: Let's just make a clean copy/translation of our travis CI setup. We can update things later. Doing it in steps will make future debugging easier.

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install poetry
+        run: |
+          pip install -U pip
+          pip install wheel
+          pip install "poetry==1.0.9"
+
+      - name: make
```
riddell-stan, line 34:
```suggestion
      - name: Build libraries used by Stan
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,64 @@
...
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install poetry
+        run: |
+          pip install -U pip
+          pip install wheel
```
riddell-stan, line 31: I'm confused about `wheel` being installed here. It's not installed in .travis.yml.

Since we're all learning this new CI configuration language, I think having a clean translation would be helpful.

riddell-stan (changes requested):

.github/workflows/main.yml
riddell-stan: This comment should point to the pybind11 segment.

riddell-stan:
One additional thing, can we discuss adding new macOS images in a separate issue?

.github/workflows/main.yml
ahartikainen: I will drop 3.9 for now. Let's wait until we get wheels pysimdjson

.github/workflows/main.yml
ahartikainen: Yes, the wheel was for pybind11 (it builds it from source).

ahartikainen:
I cleaned the .yml file.

Let's wait for pysimdjson wheel.

I also dropped Big Sur image, but we can add it in with later PR.

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  tests:
+    name: HTTPStan tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
+        python-version: [3.7, 3.8]
+      fail-fast: false
```
riddell-stan, line 18:
```suggestion
```
We do want to abort if any step fails.

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+on:
+  push:
+    branches:
+      - '**'
+    tags:
+      - '**'
+
+jobs:
+  tests:
+    name: HTTPStan tests
```
riddell-stan, line 12:
```suggestion
    name: httpstan tests
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
+        python-version: [3.7, 3.8]
+      fail-fast: false
+    steps:
+      - name: Checkout github
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
riddell-stan, line 23:
```suggestion
      - name: Check out repository
        uses: actions/checkout@v2
```

- Check out is two words when used as a verb.
- I think ``fetch-depth: 0`` is incorrect. See the documentation: "Set fetch-depth: 0 to fetch all history for all branches and tags."

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
```
riddell-stan, line 25:
```suggestion
        uses: actions/setup-python@v2
```

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
+        python-version: [3.7, 3.8]
+      fail-fast: false
+    steps:
+      - name: Checkout github
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Set up Python ${{ matrix.python-version }}
```
riddell-stan, line 24:
```suggestion
      - name: Set up Python
```
Saves 29 characters.

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+      - name: Install poetry
+        run: |
+          pip install -U pip
```
riddell-stan, line 30:
```suggestion
```
This is not part of travis-ci. If failing to upgrade pip is causing problems, let's discuss it in an issue. From my perspective, if upgrading pip is required to install httpstan, then we need to update the instructions. Fortunately, upgrading pip is not required right now.

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Build libraries used by Stan
+        run: |
+          make
+
+      - name: Install dependencies
+        run: |
+          # export test dependencies from pyproject.toml, install them
+          poetry export -f requirements.txt --dev -o requirements.txt
+          pip install -r requirements.txt
```
riddell-stan, line 41:
```suggestion
        # export test dependencies from pyproject.toml, install them
        run: poetry export -f requirements.txt --dev -o requirements.txt && pip install -r requirements.txt
```
The comment is a YAML comment and not part of the executed text

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Check code
+        run: |
+          scripts/check
```
riddell-stan, line 45:
```suggestion
        run: scripts/check
```

.github/workflows/main.yml
riddell-stan: Thanks.

Also, we don't use pybind11 from PyPI. We download and install it via `make`. And we use an older version than is available on PyPI.

riddell-stan (changes requested):
Thanks for the changes. I think we're getting close.

riddell-stan:
The commit summary needs to conform to the PyStan developer guidelines, <https://pystan-next.readthedocs.io/en/latest/contributing.html#commit-messages>. This is important because we will eventually automate the production of release notes. "ci: Switch to GitHub actions" would be better I think. Astropy also has examples: https://docs.astropy.org/en/stable/development/workflow/git_edit_workflow_examples.html#astropy-fix-example

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
+name: HTTPStan
+
+on:
+  push:
+    branches:
+      - '**'
+    tags:
+      - '**'
```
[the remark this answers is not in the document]
ahartikainen: This should work. I think that previous structure is legacy from my personal repos.

.github/workflows/main.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  tests:
+    name: HTTPStan tests
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-20.04, macos-10.15]
```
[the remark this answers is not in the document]
ahartikainen: I don't think Actions have 10.14 support. I can try.

.github/workflows/main.yml
ahartikainen: Sure

.github/workflows/main.yml
ahartikainen: Always good to save some virtual ink

.github/workflows/main.yml
ahartikainen: Yes, we can remove this

.github/workflows/main.yml
ahartikainen: I think fetch-depth is also my legacy stuff from some repo with idk some problem.

.github/workflows/main.yml
ahartikainen: Sure. (I'm not sure if there was some problem with v2 and therefore there was a reason I used v1 or maybe it is just again legacy).

.github/workflows/main.yml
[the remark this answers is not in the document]
ahartikainen: Nope, there is support for it.

riddell-stan (approved):
Great. Looks good. Thanks for doing this. I imagine we can copy most of this and use it in pystan-next.

riddell-stan:
Given that travis is broken, I merged this without waiting for any checks.

PR #499, opened 2020-11-27
feat: Compress draws and messages using LZ4

### LINKED ISSUE #485: Use fast lz4 compression for fits
For a (very) large fit, gzip fast takes 12 seconds, gzip default takes 47 seconds.

riddell-stan:
Compress draws and messages from Stan using LZ4.  Previously gzip was
used. (LZ4 is much faster than gzip.) This commit also removes
compression/decompression steps from the `load_fit` and `dump_fit`
functions. Cache functions should be as simple as possible. In
particular, they should not need to know the format of data being
stored.  That said, an indication of format remains in the filename
`jsonlines.lz4`.  This is unfortunate and should be addressed in a
future commit.

Closes #485

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -18,6 +18,8 @@ import socket
 import tempfile
 import typing
 
+import lz4.frame
+
 import httpstan.cache
 import httpstan.models
 import httpstan.services.arguments as arguments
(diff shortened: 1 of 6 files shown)


httpstan/services_stub.py
```diff
@@ -133,10 +135,16 @@ async def call(
...
                 await asyncio.sleep(0.001)
 
-    for fh in messages_files.values():
-        fh.flush()
-    httpstan.cache.dump_fit(b"".join(fh.getvalue() for fh in messages_files.values()), fit_name)
-    for fh in messages_files.values():
-        fh.close()
+    # WISHLIST: Here we compress messages after they all have arrived. Find a way to compress
+    # messages as they arrive.  Compressing messages as they arrive would use much less memory.
+    with lz4.frame.LZ4FrameCompressor() as compressor:
```
ahartikainen, line 140: The compress level by default is 0 (lowest compression).

Any idea how increasing compression changes compression time vs compression rate?

What is enough, like sending data over http vs compressing on cpu.

ahartikainen (approved):
LGTM

One typo and one question about compression level (this doesn't need to be handled in this PR if any changes are needed)

httpstan/services_stub.py
riddell-stan: I think our choice is between no compression and lowest/fastest compression. Compression slows things down. And speed is a bigger problem for us right now. We're still much slower than CmdStan.

PR #506, opened 2020-11-30
test: Exclude parts of code from coverage calculation

riddell-stan:
Exclude several sections of code from coverage calculation.  These
sections contain code that is repetitive or which is related to
debugging.


riddell-stan:
Trivial changes to comments. No code changed. Merged without review.

PR #508, opened 2020-12-01
Final adjustment of code which is ignored during coverage calculation

### LINKED ISSUE #442: Contents of .coveragerc should be in pyproject.toml
If `tomlkit` is installed, coverage can read pyproject.toml.

riddell-stan:
Adjust ignored code for coverage calculation. Completes
work started in commit 5417907.

Closes #442

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,7 +40,7 @@ jobs:
           python -m pip install dist/*.whl
 
       - name: Run tests
-        run: "python -m pytest -s -v --cov=httpstan tests"
+        run: python -m pytest -s -v --cov=httpstan tests
 
       - name: Test PyStan compatibility
         run: |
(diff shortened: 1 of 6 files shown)


.github/workflows/main.yml
```diff
@@ -40,10 +40,7 @@ jobs:
           python -m pip install dist/*.whl
 
       - name: Run tests
-        run: |
-          mkdir testdir
-          cd testdir
-          python -m pytest -s -v --cov=httpstan --cov-fail-under=93 ../tests
+        run: python -m pytest -s -v --cov=httpstan tests
```
ahartikainen, line 43: I guess this is ok. I think running from different dir than the main dir, helps in catching the  relative path / imports issues if there are any.

ahartikainen (approved):

PR #509, opened 2020-12-03
feat: Add log_prob_grad method and endpoint

mjcarter95:
Given a model name, associated data and unconstrained parameters,
the log_prob_grad endpoint will calculate and return the gradient
of the unconstrained parameters.

This feature is accompanied by a test: the log_prob_grad method is
validated on a simple Gaussian example, the test compares the output
of the endpoint against an analytical calculation of the gradient.

### DIFF
--- a/httpstan/stan_services.cpp
+++ b/httpstan/stan_services.cpp
@@ -165,6 +165,43 @@ double log_prob(py::dict data,
 }
 
 // See exported docstring
+std::vector<double> log_prob_grad(py::dict data,
+                                  const std::vector<double>& unconstrained_parameters,
+                                  bool adjust_transform) {
+  std::vector<double> gradient;
+  stan::io::array_var_context &var_context = new_array_var_context(data);
+  // random_seed, the second argument, is unused but the function requires it.
+  stan::model::model_base &model = new_model(var_context, (unsigned int)1, &std::cout);
+  if(unconstrained_parameters.size() != model.num_params_r()) {
+    throw std::runtime_error("The number of parameters does not match the number of unconstrained parameters in the model.");
+  }
+  // The params_r parameter is incorrectly declared as non-const in Stan C++.
+  // Unconstrained_parameters are cast from const to non-const below, as required by Stan (see model_base.hpp).
+  std::vector<double>& params_r = const_cast<std::vector<double>&>(unconstrained_parameters);
+  // calculate gradient
+  std::exception_ptr p;
+  std::vector<int> params_i(model.num_params_i(), 0);
+  try {
+    // params_i, the third argument, is unused but the function requires it (see model_base.hpp).
+    if(adjust_transform) {
+      stan::model::log_prob_grad<true, true>(model, params_r, params_i, gradient, &std::cout);
+    } else {
+      stan::model::log_prob_grad<true, false>(model, params_r, params_i, gradient, &std::cout);
+    }
+  } catch (std::exception& ex) {
+    p = std::current_exception();
+  }
+
+  delete &model;
+  delete &var_context;
+
+  if (p)
+    std::rethrow_exception(p);
+
+  return gradient;
+}
+
+// See exported docstring
 int hmc_nuts_diag_e_adapt_wrapper(std::string socket_filename, py::dict data, py::dict init, int random_seed,
                                   int chain, double init_radius, int num_warmup, int num_samples, int num_thin,
                                   bool save_warmup, int refresh, double stepsize, double stepsize_jitter,
@@ -222,6 +259,8 @@ PYBIND11_MODULE(stan_services, m) {
   m.def("dims", &dims, py::arg("data"), "Call the ``get_dims`` method of the model.");
   m.def("log_prob", &log_prob, py::arg("data"), py::arg("unconstrained_parameters"), py::arg("adjust_transform"),
         "Call the ``log_prob`` method of the model.");
+  m.def("log_prob_grad", &log_prob_grad, py::arg("data"), py::arg("unconstrained_parameters"), py::arg("adjust_transform"),
+        "Call stan::model::log_prob_grad");
   m.def("hmc_nuts_diag_e_adapt_wrapper", &hmc_nuts_diag_e_adapt_wrapper, py::arg("socket_filename"), py::arg("data"),
         py::arg("init"), py::arg("random_seed"), py::arg("chain"), py::arg("init_radius"), py::arg("num_warmup"),
         py::arg("num_samples"), py::arg("num_thin"), py::arg("save_warmup"), py::arg("refresh"), py::arg("stepsize"),
--- a/httpstan/views.py
+++ b/httpstan/views.py
@@ -634,3 +634,81 @@ async def handle_log_prob(request: aiohttp.web.Request) -> aiohttp.web.Response:
         logger.critical(message)
         return aiohttp.web.json_response(_make_error(message, status=status), status=status)
     return aiohttp.web.json_response({"log_prob": lp}, status=200)
+
+
+async def handle_log_prob_grad(request: aiohttp.web.Request) -> aiohttp.web.Response:
+    """Calculate the gradient.
+
+    ---
+    post:
+      summary: Return the gradient of the log posterior evaluated at the unconstrained parameters.
+      description: >-
+        Returns the output of Stan C++ `stan::model::log_prob_grad`.
+      consumes:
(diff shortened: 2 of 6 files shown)


mjcarter95:
The tests are failing during the "install dependencies" stage - @riddell-stan @ahartikainen any ideas why?

riddell-stan:
"multidict==5.1.0" should be satisfiable. Perhaps it's a PyPI cache issue. (5.1.0 was released 5 hours ago.) I restarted the job.

mjcarter95:
The tests run and pass on my fork of HTTPStan - please could you re-run the job again when you get chance? I tried to do this but don't see the option on the main repo.

ahartikainen:
I think there is a bug in `pip`.

@riddell-stan do we wait until actions updates the version

https://pypi.org/project/pip/#history

riddell-stan:
Let's wait until `pip` fixes itself.

I'll review this because I want to make sure we get the function names right. (httpstan, as a thin "shim" on top of Stan C++, should use Stan's function and method names.) I may not have time to review this until the weekend though.

mjcarter95:
I've been following PyStan2 naming conventions so far, but can easily rename to be in line with Stan - with that in mind, should the `log_prob` method/endpoint be updated to `log_prob_propto`?

riddell-stan:
@mjcarter95 re: log_prob_propto, yes, I think we do want to make a change.

A related question, is there any use for the `log_prob` method of the model? It comes right before `write_array` in the generated code.

mjcarter95:
@riddell-stan See below. `log_prob_propto` calls the `log_prob` method of the model - I think we could mirror this behaviour in `httpstan` and bypass calls to `stan::model`, what do you think?

```
    double log_prob_propto(const M& model,
                           std::vector<double>& params_r,
                           std::vector<int>& params_i,
                           std::ostream* msgs = 0) {
      using stan::math::var;
      using std::vector;
      vector<var> ad_params_r;
      ad_params_r.reserve(model.num_params_r());
      for (size_t i = 0; i < model.num_params_r(); ++i)
        ad_params_r.push_back(params_r[i]);
      try {
        double lp
          = model.template log_prob<true, jacobian_adjust_transform>
          (ad_params_r, params_i, msgs).val();
        stan::math::recover_memory();
        return lp;
      } catch (std::exception &ex) {
        stan::math::recover_memory();
        throw;
      }
    }
```

riddell-stan:
I think I understand what you're saying. Sounds good to me.

The general concern here is that httpstan's official goal is providing 
access to functions in stan::services and model methods -- and to no 
other functions.

Clearly we have to make an exception for the grad log prob. But one 
exception is better than two.

mjcarter95:
Neat. I have a few other bits of work to complete then I'll have a go at calling the `log_prob` method in the model and submit a "fix". I'll also tweak the naming convention in this PR to follow Stan's and not PyStan's (i.e. `log_prob_grad`, not `grad_log_prob`).

mjcarter95:
Renamed the endpoint, tests etc. to `log_prob_grad` in the latest commit to reflect the name of the stan::model function we call. Should this PR be closed and replaced by a PR from mjcarter95:feature/grad_log_prob?

riddell-stan:
No need to change the branch name.

httpstan/views.py
```diff
@@ -634,3 +634,81 @@ async def handle_log_prob(request: aiohttp.web.Request) -> aiohttp.web.Response:
...
         return aiohttp.web.json_response(_make_error(message, status=status), status=status)
     return aiohttp.web.json_response({"log_prob": lp}, status=200)
+
+
+async def handle_log_prob_grad(request: aiohttp.web.Request) -> aiohttp.web.Response:
+    """Calculate the gradient.
+
+    ---
+    post:
+      summary: Return the gradient of the unconstrained parameters.
```
riddell-stan, line 644: Is this the right language? Is it the gradient of log_prob evaluated at the provided parameters?

httpstan/views.py
```diff
@@ -634,3 +634,81 @@ async def handle_log_prob(request: aiohttp.web.Request) -> aiohttp.web.Response:
...
+          description: >-
+              Boolean to control whether we apply a Jacobian adjust transform.
+          required: true
+          schema:
+            type: array
+            items:
+              type: number
+      responses:
+        "200":
+          description: Gradient of the unconstrained parameters.
```
riddell-stan, line 681: See above.

riddell-stan (changes requested):
In general, looks good. The docstrings could be a bit more verbose (or different).

mjcarter95:
Updated the docstrings and rebased on main in latest commit (this fixed the conflicts in the previous commit).

httpstan/stan_services.cpp
```diff
@@ -164,6 +164,34 @@ double log_prob(py::dict data,
...
+  // random_seed, the second argument, is unused but the function requires it.
+  stan::model::model_base &model = new_model(var_context, (unsigned int)1, &std::cout);
+  if(unconstrained_parameters.size() != model.num_params_r()) {
+    throw std::runtime_error("The number of parameters does not match the number of unconstrained parameters in the model.");
+  }
+  // copy unconstrained_parameters -> params_r
+  std::vector<double> params_r(unconstrained_parameters.size());
+  std::memcpy(params_r.data(), unconstrained_parameters.data(), unconstrained_parameters.size()*sizeof(double));
+  // calculate logprob
+  std::vector<int> params_i(model.num_params_i(), 0);
```
riddell-stan, line 181: Two questions:
1. Is the copy into params_r necessary?
2. What about params_i? What if the model has integer parameters?

mjcarter95: responses below

> 1. Is the copy into params_r necessary?
If the unconstrained_parameters argument is defined as `const std::vector<double>& unconstrained_parameters` then the copy to params_r is necessary, defining as `std::vector<double> unconstrained_parameters` means the copy isn't necessary. I'm not sure which is preferred?

> 2. What about params_i? What if the model has integer parameters?
The model base class (see https://github.com/stan-dev/stan/blob/develop/src/stan/model/model_base.hpp) suggests that `params_i` are ignored, so I don't think it's necessary to have an argument for them.

riddell-stan:
Let's use the non-const std::vector<double> then. That seems to match 
what is used in the {model_name}.hpp file. And doing so saves several 
lines of code.

A note about params_i being ignored would be nice. Perhaps say "see 
model_base.hpp".

mjcarter95:
Cool. Is it worth also updating `log_prob` to use non-const in the same PR? Shame that we just closed the log_prob fix PR.

riddell-stan:
Sure. Open up a new PR.

Also, I fixed the wheel building yesterday in httpstan-wheels. So we can 
make releases now.

Making a release requires a little bit of effort, so if `write_array` 
can be done before I make a release, that would be ideal.

mjcarter95:
Amended `log_prob_grad` in the latest commit, will fix/open PR for `log_prob` now. I'll work on adding `write_array` and `transform_inits` in an hour or so.

riddell-stan:
Sorry to keep digging into this. My C++ is somewhat rusty. It seems like we can use const_cast, https://en.cppreference.com/w/cpp/language/const_cast here.

Our `const vector<double> &params_r` is correct. It's the stanc generated code which is incorrect (it should have const), right?

Perhaps a better compromise would be to (unsafely?) remove the const, with const_cast and then explain in an inline comment why we are doing it?

mjcarter95:
My C++ is a bit rusty as well, however, agreed re stanc: I think the generated code would ideally have const such that the argument is read only. With that in mind, I wonder if it's best to keep the copy given that `const_cast` results in undefined behaviour?

riddell-stan:
Fair enough. Can we at least find a way to copy a const vector to a non-const vector in a single line (i.e., without memcpy)? There has got to be a way.

mjcarter95:
I've done a bit more reading and seems like the undefined behaviour only occurs when _modifying_ a value that was initially const, so it should be fine to use for what we want given that we are just reading. I'll modify and add a line to explain what we are doing soon.

mjcarter95:
Updated to use `const_cast` in the latest commit.

mjcarter95:
Added try/except in case gradient evaluation fails.

riddell-stan (approved):

riddell-stan:
LGTM. Is this ready to be merged?

mjcarter95:
If you're happy with it, can't think of anything else to add.

PR #510, opened 2020-12-05
fix: Call log_prob method of model class

mjcarter95:
Previously, the log_prob endpoint called the
log_prob_propto method in the stan::model namespace.
This fix calls the log_prob method of the model  class instead.

### DIFF
--- a/httpstan/stan_services.cpp
+++ b/httpstan/stan_services.cpp
@@ -125,6 +125,7 @@ std::vector<std::vector<size_t>> dims(py::dict data) {
 double log_prob(py::dict data,
                 const std::vector<double>& unconstrained_parameters,
                 bool adjust_transform) {
+  using stan::math::var;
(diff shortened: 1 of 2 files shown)


mjcarter95:
Should I rename the endpoint to `log_prob_propto` to reflect the naming in Stan? It's effectively the same function.

ahartikainen:
That sounds reasonable.

riddell-stan:
Shouldn't we keep the endpoint as `log_prob` since it's calling the method of that name?

mjcarter95:
Latest push amends an error in the commit message.

On renaming: looking back at our previous discussions, we spoke about naming endpoints after the functions/methods that they call (rather than their Stan equivalents) so keeping as `log_prob` makes sense.

riddell-stan (approved):
LGTM.

riddell-stan:
HTTP Status 201 is "Created" which doesn't apply here. 200 is good.

PR #512, opened 2020-12-07
fix: Remove unnecessary memcpy and reorder logprob

mjcarter95:
Removes unncessary calls to the memcpy (we already copied from a vector of
doubles to a vector of stan::math::var). Slight reordering to return the
log probability after deleting model and var_context variables.


mjcarter95:
Push corrects an error I missed in the commit message.

@riddell-stan Given that we copy from a vector of doubles to a vector of stan::math::var (the type required by log_prob method of {model_name}.hpp), I didn't see the need to use `const_cast` here. Do you think this needs an inline comment to say that we copy from one vector to another to ensure correct types?

riddell-stan (approved):
This looks good. Thanks for paying attention to the details.

PR #513, opened 2020-12-08
feat: Add write_array method and endpoint

mjcarter95:
Given a model name, associated data and unconstrained parameters,
the write_array endpoint will transform the sequence of unconstrained
parameters to their defined support, optionally including transformed
parameters and generated quantities.

### DIFF
--- a/httpstan/views.py
+++ b/httpstan/views.py
@@ -712,3 +712,97 @@ async def handle_log_prob_grad(request: aiohttp.web.Request) -> aiohttp.web.Resp
         logger.critical(message)
         return aiohttp.web.json_response(_make_error(message, status=status), status=status)
     return aiohttp.web.json_response({"log_prob_grad": gradient}, status=200)
+
+
+async def handle_write_array(request: aiohttp.web.Request) -> aiohttp.web.Response:
+    """Constrain parameters.
+
+    Transform a sequence of unconstrained parameters to their defined support,
+    optionally including transformed parameters and generated quantities.
+
+    Note: This endpoint exposes the ``write_array`` method of the model class.
+    This method is utilised in the ``constrain_pars`` function in PyStan and RStan.
+
+    ---
+    post:
+      summary: Return a sequence of unconstrained parameters.
+      description: >-
+        Returns the output of Stan C++ ``write_array`` model class method.
+      consumes:
+        - application/json
+      produces:
+        - application/json
+      parameters:
+        - name: model_id
+          in: path
+          description: ID of Stan model to use
+          required: true
+          type: string
+        - in: body
+          name: data
+          description: >-
+              Data for the Stan Model.
+          required: true
+          schema: Data
+        - in: body
+          name: unconstrained_parameters
+          description: >-
+              Sequence of unconstrained parameters.
+          required: true
+          schema:
+            type: array
+            items:
+              type: number
+        - in: body
+          name: include_tparams
+          description: >-
+              Boolean to control whether we include transformed parameters.
+          required: true
+          schema:
+            type: boolean
+        - in: body
+          name: include_gqs
+          description: >-
+              Boolean to control whether we include generated quantities.
+          required: true
+          schema:
+            type: boolean
+      responses:
+        "200":
+          description:
+              Sequence of constrained parameters, optionally including transformed parameters
+              and generated quantities.
+          schema:
+            type: object
+            properties:
+              params_r_constrained:
+                type: array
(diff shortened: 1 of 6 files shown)


mjcarter95:
A few points I'd appreciate your input on (@riddell-stan, @ahartikainen)

1. I'm not sure how to test this endpoint, any ideas?
2. I'm met with an error when running the mypy tests, but not sure why it's there: can you see what's causing it?

httpstan/views.py
```diff
@@ -712,3 +712,96 @@ async def handle_log_prob_grad(request: aiohttp.web.Request) -> aiohttp.web.Resp
...
+    try:
+        services_module = httpstan.models.import_services_extension_module(model_name)
+    except KeyError:
+        message, status = f"Model `{model_name}` not found.", 404
+        return aiohttp.web.json_response(_make_error(message, status=status), status=status)
+
+    try:
+        params_r_constrained = services_module.write_array(
+            data, unconstrained_parameters, include_tparams, include_gqs
+        )  # type: ignore
```
ahartikainen, line 802: I think this fails due to multiline

mjcarter95: Ah yes, that fixed it - cheers!

mjcarter95:
Cheers @ahartikainen, mypy error resolved. Now seeing a coverage error (due to lack of testing for the write_array endpoint): any ideas on how to test this?

ahartikainen:
For simple model

```
parameters {
    real y;
}
```

init of `x` -> `x`

And for 

```
parameters {
    real<lower=0> y;
}
```
init of `x` -> `log(x)`

So maybe test something along those lines?

mjcarter95:
Cool. Added a test that covers the cases you mentioned and also checks transformed parameters and generated quantities.

httpstan/views.py
```diff
@@ -712,3 +712,94 @@ async def handle_log_prob_grad(request: aiohttp.web.Request) -> aiohttp.web.Resp
         logger.critical(message)
         return aiohttp.web.json_response(_make_error(message, status=status), status=status)
     return aiohttp.web.json_response({"log_prob_grad": gradient}, status=200)
+
+
+async def handle_write_array(request: aiohttp.web.Request) -> aiohttp.web.Response:
+    """Constrain parameters.
+
+    Transform a sequence of unconstrained parameters to their defined support,
+    optinally including transformed parameters and generated quantities.
```
ahartikainen, line 721:
```suggestion
    optionally including transformed parameters and generated quantities.
```

ahartikainen (approved):
Small typo, but otherwise looks good.

mjcarter95:
Hopefully will have a PR for `transform_inits` (to unconstrain parameters) tomorrow. Also noticed a few errors in the `log_prob` and `log_prob_grad` docs (apologies!), I'll submit a PR to correct these tonight if I get chance.

riddell-stan:
If you have a moment, please add to the docstrings a reminder for future 
readers about why these particular functions/methods are being exposed. 
I'm concerned that people may not immediately notice the connection 
between these Stan functions and their use (under different names) in 
PyStan and RStan.

mjcarter95:
> If you have a moment, please add to the docstrings a reminder for future readers about why these particular functions/methods are being exposed. I'm concerned that people may not immediately notice the connection between these Stan functions and their use (under different names) in PyStan and RStan.

I can do this whilst correcting the `log_prob` and `log_prob_grad` docstrings, will submit in a separate PR.

riddell-stan:
This looks good. Can we discuss the parameter names? For example, where does `include_tparams` come from? I see an ungainly `emit_transformed_parameters__` in the generated C++ code. Are you looking at a different write_array than I am?

riddell-stan:
Yes, I was thinking of `write_array` and `transform_inits`. I'm more worried about confusion with these functions.

mjcarter95:
> This looks good. Can we discuss the parameter names? For example, where does `include_tparams` come from? I see an ungainly `emit_transformed_parameters__` in the stanc-generated C++ code. Are you looking at a different write_array than I am?

I'm looking at the definition in model_base.hpp (see https://github.com/stan-dev/stan/blob/develop/src/stan/model/model_base.hpp).

PS on the `transform_inits` I'm a bit confused of how to work with `stan::io::var_context`, I suspect I'm overlooking something, but not sure I'll have time to revisit exposing this function until the weekend.

riddell-stan:
That sounds like a good place. It's odd stanc uses different names.

PR #514, opened 2020-12-09
docs: Correct log_prob and log_prob_grad docstring

mjcarter95:
Minor changes to the docstrings of log_prob and log_prob_grad.

### DIFF
--- a/httpstan/views.py
+++ b/httpstan/views.py
@@ -590,16 +590,16 @@ async def handle_log_prob(request: aiohttp.web.Request) -> aiohttp.web.Response:
               Unconstrained parameters to calculate log probability for.
           required: true
           schema:
-            type: boolean
+            type: array
+            items:
+              type: number
         - in: body
           name: adjust_transform
           description: >-
               Boolean to control whether we apply a Jacobian adjust transform.
-          required: true
+          required: false
           schema:
-            type: array
-            items:
-              type: number
+            type: boolean
       responses:
         "200":
           description: Log probability of the unconstrained parameters.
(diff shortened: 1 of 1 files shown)


riddell-stan:
Just had a quick look. One note:

Perhaps something is missing? Maybe we need another test. If the user does not provide adjust_transform, does this not raise an exception?

If adjust_transform needs to be specified in Stan C++, we should require that it be given here.

mjcarter95:
Hmm.. the schemas for `log_prob` and `log_prob_grad` should (theoretically) validate the request and set the default boolean parameters to true if they aren't specified by the user - this is why I changed required to false in the documents: should we base the docs on the behaviour of the Stan C++ method?

Also, I noticed an error in the `log_prob_grad` schema (doh!) we should validate `include_tparams` and `include_gqs`, not `adjust_transform`.

```
class ShowLogProbGradRequest(marshmallow.Schema):
    """Schema for log_prob_grad request."""

    data = fields.Nested(Data(), missing={})
    unconstrained_parameters = fields.List(fields.Float(), required=True)
    adjust_transform = fields.Boolean(missing=True)
```

mjcarter95:
Hmm.. the schemas for `log_prob` and `log_prob_grad` should (theoretically) validate the request and set the default boolean parameters to true if they aren't specified by the user - this is why I changed required to false in the documents: should we base the docs on the behaviour of the Stan C++ method?

Additionally, is it worth adding a test when `adjust_transform` is true? We currently only test when this is false.

riddell-stan:
@mjcarter95 I'm so sorry. I forgot that `missing` is marshmallow's way of specifying defaults. (I thought it implied that a missing field was allowed.)

No need to add a test, I think. This is a feature for expert users. I trust they'll report problems.

riddell-stan (approved):

PR #517, opened 2020-12-14
Update poetry and associated docs

riddell-stan:
I believe poetry has solved the problem which #466 addressed.


riddell-stan:
Merged without review as more than 48 hours have passed since a review was requested.

PR #518, opened 2020-12-18
feat: Add transform_inits method and endpoint

mjcarter95:
Given a model name, associated data, and constrained parameter
values and their context, the transformed_inits endpoint will
unconstrain the constrained parameters from their defined
support to unconstrained space.

### DIFF
--- a/httpstan/stan_services.cpp
+++ b/httpstan/stan_services.cpp
@@ -6,6 +6,7 @@
 #include <stan/callbacks/stream_logger.hpp>
 #include <stan/callbacks/writer.hpp>
 #include <stan/io/array_var_context.hpp>
+#include <stan/io/var_context.hpp>
 #include <stan/model/model_base.hpp>
 #include <stan/services/sample/hmc_nuts_diag_e_adapt.hpp>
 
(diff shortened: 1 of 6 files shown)


riddell-stan:
Is there some kind of type error? `io::array_var_context` subclasses `io::var_context`, right?

mjcarter95:
Yes `io::array_var_context` subclasses `io::var_context`. `transform_inits` (see below) only accepts `io::var_context`, but I'm not sure how to initialise this (I intended on using methods in the cmdstan::json namespace but can't access these.

```
virtual void transform_inits(const io::var_context& context,
                               Eigen::VectorXd& params_r,
                               std::ostream* msgs) const = 0;
```

riddell-stan:
Can you declare var_context to be an ``stan:io::var_context``? Like:
```
stan::io::var_context &var_context = new_array_var_context(data);
```

mjcarter95:
Well that was easy - code now compiles, cheers.

httpstan/stan_services.cpp
```diff
@@ -233,6 +233,34 @@ std::vector<double> write_array(py::dict data,
...
+  stan::io::array_var_context &var_context = new_array_var_context(data);
+  // random_seed, the second argument, is unused but the function requires it.
+  stan::model::model_base &model = new_model(var_context, (unsigned int)1, &std::cout);
+  stan::io::array_var_context &param_var_context = new_array_var_context(constrained_parameters);
+  // unconstrain parameters from their defined support
+  std::exception_ptr p;
+  std::vector<int> params_i(model.num_params_i(), 0);
+  try {
+    // params_i, the third argument, is unused but the function requires it (see model_base.hpp).
+    model.transform_inits(param_var_context, params_r_unconstrained, params_i, &std::cout);
```
riddell-stan, line 249: If transform_inits can throw an exception, a comment reminding us why might be helpful here. (I find myself wondering the same thing for write_array and log_prob (with unconstrained params), actually.)

mjcarter95:
Added the `transform_inits` method, endpoint and test in latest commit

httpstan/stan_services.cpp
mjcarter95: Hmm.. `log_prob` throws an exception if the evaluation fails (you can see this behaviour in Stan's initialisation method). I'm not 100% sure why write_array/transform_inits would fail, but have seen this mentioned on the discourse/main Stan repo which is why I added the try just to be safe.

mjcarter95:
Not sure why coverage fell below 95% here?

riddell-stan:
I'll remove the coverage checks. It's clear to me that coverage checks are not helping us develop this piece of software.

I've been giving some thought to the exception handling blocks for all these new functions. I think there's a cleaner way of doing things. Eventually I'll have time to test out my idea.

In the meantime, however, I think at least one of these new functions needs a test that triggers an exception---given that we are adding these C++ lines which deal with potential exceptions. log_prob might be the easiest one.

Sorry to add this to your queue. (We do have other tests which verify that C++ exceptions are handled.)

mjcarter95:
I'll add a test to `log_prob`, are you happy for the change to form part of this PR? Or should I submit a separate PR?

PS I have `grad_log_prob` and `constrain_pars` exposed in PyStan-Next, once we've ironed this out I'll add `unconstrain_pars` and tests

riddell-stan (approved):
This all looks very good. Thank you!

If the only issue is coverage, rebasing this on main should fix that.

mjcarter95:
Rebased on main in latest push.

PR #520, opened 2020-12-23
Increase pathlib.Path use

ahartikainen:
Switch to use pathlib.Path as the main way to handle path information.

### DIFF
--- a/httpstan/cache.py
+++ b/httpstan/cache.py
@@ -3,11 +3,10 @@
 Functions in this module manage the Stan model cache and related caches.
 """
 import logging
-import os
-import pathlib
 import shutil
 import typing
 from importlib.machinery import EXTENSION_SUFFIXES
+from pathlib import Path
 
 import appdirs
 
@@ -16,11 +15,23 @@ import httpstan
 logger = logging.getLogger("httpstan")
 
 
-def model_directory(model_name: str) -> str:
+def cache_directory() -> Path:
+    """Get httpstan cache path."""
+    return Path(appdirs.user_cache_dir("httpstan", version=httpstan.__version__))
+
+
+def model_directory(model_name: str) -> Path:
     """Get the path to a model's directory. Directory may not exist."""
-    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
     model_id = model_name.split("/")[1]
-    return os.path.join(cache_path, "models", model_id)
+    return cache_directory() / "models" / model_id
+
+
+def fit_path(fit_name: str) -> Path:
+    """Get the path to a fit file. File may not exist."""
+    # fit_name structure: cache / models / model_id / fit_id
+    fit_directory, fit_id = fit_name.rsplit("/", maxsplit=1)
+    fit_filename = fit_id + ".jsonlines.lz4"
+    return cache_directory() / fit_directory / fit_filename
 
 
 def delete_model_directory(model_name: str) -> None:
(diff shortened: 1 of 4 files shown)


riddell-stan:
Is this ready to be reviewed?

ahartikainen:
Yes (I forgot the PR 👍 )

httpstan/cache.py
```diff
@@ -94,12 +96,10 @@ def dump_fit(fit_bytes: bytes, name: str) -> None:
         name: Stan fit name
         fit_bytes: LZ4-compressed messages associated with Stan fit.
     """
-    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
     # fits are stored under their "parent" models
-    fits_path = os.path.join(*([cache_path] + name.split("/")[:-1]))
-    fit_filename = os.path.join(fits_path, f'{name.split("/")[-1]}.jsonlines.lz4')
-    os.makedirs(fits_path, exist_ok=True)
-    with open(fit_filename, mode="wb") as fh:
+    fit_filename = cache_directory() / (name + ".jsonlines.lz4")
```
riddell-stan, line 100: The previous version was a bit more explicit ("Explicit is better than implicit."). Also, can this use model_directory instead? In theory, the function should not need to know the cache directory. (The original was unskillful.)

httpstan/cache.py
```diff
@@ -113,12 +113,10 @@ def load_fit(name: str) -> bytes:
     Returns
         LZ4-compressed messages associated with Stan fit.
     """
-    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
     # fits are stored under their "parent" models
-    fits_path = os.path.join(*([cache_path] + name.split("/")[:-1]))
-    fit_filename = os.path.join(fits_path, f'{name.split("/")[-1]}.jsonlines.lz4')
+    fit_filename = cache_directory() / (name + ".jsonlines.lz4")
```
riddell-stan, line 117: Maybe we need a fits_directory function to clean this all up?

httpstan/cache.py
```diff
@@ -130,7 +128,5 @@ def delete_fit(name: str) -> None:
     Arguments:
         name: Stan fit name
     """
-    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
-    fits_path = os.path.join(*([cache_path] + name.split("/")[:-1]))
-    fit_id = name.split("/")[-1]
-    pathlib.Path(os.path.join(fits_path, f"{fit_id}.jsonlines.lz4")).unlink()
+    fit_filename = cache_directory() / (name + ".jsonlines.lz4")
```
riddell-stan, line 131: Another use for fits_directory

riddell-stan (commented):
This looks good. Thanks for doing this.

How about adding a `fit_directory` function (which resembles `model_directory`). I think it will make the code more readable.

riddell-stan:
@ahartikainen Any thoughts about my comments?

ahartikainen:
Yes, I will fix the code when I have time. (I think this can go in in later release, no hurry right?)

httpstan/cache.py
```diff
@@ -16,11 +15,23 @@
...
+
+def model_directory(model_name: str) -> Path:
     """Get the path to a model's directory. Directory may not exist."""
-    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
     model_id = model_name.split("/")[1]
-    return os.path.join(cache_path, "models", model_id)
+    return cache_directory() / "models" / model_id
+
+
+def fit_filepath(fit_name: str) -> Path:
```
riddell-stan, line 29: I'd stick with `fit_path` ("filepath" almost never occurs in the Python documentation; I think people have settled on "path" as the term)

httpstan/cache.py
```diff
@@ -94,12 +104,10 @@ def dump_fit(fit_bytes: bytes, name: str) -> None:
         name: Stan fit name
         fit_bytes: LZ4-compressed messages associated with Stan fit.
     """
-    cache_path = appdirs.user_cache_dir("httpstan", version=httpstan.__version__)
     # fits are stored under their "parent" models
-    fits_path = os.path.join(*([cache_path] + name.split("/")[:-1]))
-    fit_filename = os.path.join(fits_path, f'{name.split("/")[-1]}.jsonlines.lz4')
-    os.makedirs(fits_path, exist_ok=True)
-    with open(fit_filename, mode="wb") as fh:
+    fit_filename = fit_filepath(name)
```
riddell-stan, line 108: This should be `fit_path` not `fit_filename`, right?

httpstan/cache.py
ahartikainen: What do you think about the function? I tried to make it a bit more explicit, but I was not sure which one is cleaner solution

Split fit to model_id, fit_id and then call model_directory or just get cache path.

httpstan/cache.py
ahartikainen: Yes, fit_path sounds better

httpstan/cache.py
riddell-stan: I don't have any objection to it. It's a tiny bit odd that we have model_directory but not fit_directory. On the other hand, if we never refer to fit_directory directly, there's no need for a function returning it.

riddell-stan (approved):
This looks great. Thank you!

This just needs to be rebased now.

PR #524, opened 2021-01-20
feat: Update Stan to 2.26.0, bump version

riddell-stan:
Update Stan to 2.26.0.


ahartikainen (approved):
LGTM

What do you think we if could create a small table that lists httpstan/stan/stan-math versions? (E.g. in wiki)

riddell-stan:
Sounds like an interesting idea. Maybe open an issue and describe the idea in greater detail?

PR #530, opened 2021-02-11
fix: Place libs required by Stan Math in include

### LINKED ISSUE #469: Organization of `httpstan/include` is confusing
`httpstan/include` is poorly organized.

```
── include
│   ├── google
│   ├── lib
│   │   ├── boost_1.72.0
│   │   ├── eigen_3.3.7
│   │   ├── sundials_5.2.0
│   │   └── tbb_2019_U8
│   ├── pybind11
│   └── stan
```
Header files from `google`, `stan`, and `pybind` are treated differently than the header files in `include/lib`. The structure of `lib` is taken from `stan/math` so we cannot change that without creating more confusion (although we might move it to `stan_math/lib` to make it clear we are not responsible for its organization).

We could also rename the `INCLUDES` variable name in the makefile to something like `HEADER_FILES` so it is not confused with the Python `include_dirs`.

riddell-stan:
refactor: Consistently name source code directories

Place source code directories in `httpstan/include` in a consistent way.
Follows the convention familiar from Unix-like systems. Users should
only need to add `httpstan/include` as a search path for third-party
source files. Previously many more search paths needed to be specified.

Closes #469

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,7 @@ RAPIDJSON_VERSION := 1.1.0
 STAN_VERSION := 2.26.1
 STANC_VERSION := 2.26.1
 MATH_VERSION := 4.0.1
+# NOTE: boost, eigen, sundials, and tbb versions must match those found in Stan Math
 BOOST_VERSION := 1.72.0
 EIGEN_VERSION := 3.3.9
 SUNDIALS_VERSION := 5.6.1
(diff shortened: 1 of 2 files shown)


riddell-stan:
This needs a bit more work. It's odd that we don't have the version number for stan. And we should have a separate directory for math.

riddell-stan:
This is ready for review.

A side benefit is that this reduces the size of the wheel since we avoid including unrelated files.

ahartikainen (approved):
LGTM


Do we want to manually add version to Stan folder?

riddell-stan:
I don't think there are any benefits to adding/vendoring the Stan library. And there would be significant costs. It makes cloning the repository much slower.

PR #541, opened 2021-02-21
fix: Use asyncio.get_running_loop everywhere

riddell-stan:
Use asyncio.get_running_loop everywhere. Elsewhere, in a similar call,
we use asyncio.get_running_loop. This commit eliminates the
inconsistency.  asyncio.get_running_loop is the function we want. It
raises an exception if an event loop does not exist.


riddell-stan:

PR #544, opened 2021-02-25
fix: Get version from importlib

riddell-stan:
Use importlib to get version. Same approach used in PyStan.


ahartikainen (approved):

riddell-stan:
We need the version for the docs. We can't get the version without installing httpstan. We can't build and install httpstan on the readthedocs server (too resource intensive).

Perhaps there's a way to get the version from git and give it to readthedocs.

riddell-stan:
This change requires dropping support for Python 3.7. Best to revisit this in 2022 after Python 3.10 is out.

PR #551, opened 2021-03-09
refactor: Use method names as they appear in Stan

riddell-stan:
Use method names as they appear in the generated Stan C++ code.
For example, the method `get_dims` should be called `get_dims`
everywhere. This will avoid needless confusion.


riddell-stan:

PR #556, opened 2021-03-17
docs: Add C++ compiler version requirement

### LINKED ISSUE #501: docs: Full list of supported platforms and architectures missing
We do not have a list of supported platforms and architectures. I believe it's

Supported platforms:

- x86-64 Ubuntu Linux 20.04
- x86-64 macOS with macOS 11.14

Supported Python versions :

- CPython 3.7
- CPython 3.8
- CPython 3.9

Once we upgrade pybind11 to the current version I believe we can support PyPy7 as well.

Platforms to evaluate:
- Alpine Linux
- Arch Linux
- Centos
- Debian (the next one)
- macOS 11.15

riddell-stan:
Indicate which C++ compilers we know work.

Closes #501

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,6 +31,7 @@ Requirements
 
 - Python version 3.7 or higher.
 - macOS or Linux.
+- C++ compiler: gcc ≥9.0 or clang ≥10.0.
 
 Background
 ==========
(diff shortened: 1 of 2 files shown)


riddell-stan:

PR #558, opened 2021-03-17
fix: Fix parallel make on macOS

### LINKED ISSUE #557: make failure on macos
```
clang: error: no such file or directory: 'libtbbmalloc.dylib'
```

riddell-stan:
Add rule to allow macOS to use parallel make.

Closes #557


riddell-stan:

PR #561, opened 2021-03-28
fix: Require multiprocessing to use "fork"

riddell-stan:
Require multiprocessing to use the "fork" method for starting
new processes. This is the only version httpstan currently works
with. (macOS recently switched to "spawn" as the default.)

This follows the recommendation in the Python `multiprocessing`
documentation:

> A library which wants to use a particular start method should probably
> use get_context() to avoid interfering with the choice of the library
> user.

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -12,6 +12,7 @@ import concurrent.futures
 import functools
 import io
 import logging
+import multiprocessing as mp
 import os
 import select
 import socket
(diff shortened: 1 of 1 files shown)


riddell-stan:
This should address the problem in stan-dev/pystan#247

ahartikainen (approved):
This looks good.

Should we add some simple python script with basic import + model compilation + sampling which is called with `python simple.py`?

I think pytest handles the `__main__` namespace issue.

riddell-stan:
re: "python script with basic import + model compilation + sampling" Are you thinking this would be useful documentation for people who aren't familiar with how REST APIs work?

I think the README example with `curl` is reasonably complete.

Of course, if you see any specific weaknesses in the documentation, feel free to open an issue.

ahartikainen:
Just for having a "test" (outside pytest) that works. Maybe pystan is a better place for it?

riddell-stan:
We have such a test in GitHub Actions: https://github.com/stan-dev/httpstan/blob/a1df6f3cf849431a015e20c218fca03494568e36/.github/workflows/main.yml#L48

PR #565, opened 2021-04-27
fix: Use established idiom for deleting fits

riddell-stan:
Use the already established (Pythonic) idiom for deleting fits.
Raise a KeyError if the to-be-deleted fit does not exist.


ahartikainen (approved):
Looks good.

PR #579, opened 2021-10-02
chore: Update dependencies

riddell-stan:
Use more recent versions of packages on which httpstan depends.


riddell-stan:

PR #581, opened 2021-10-10
ci: Test Python 3.10

riddell-stan:
Add Python 3.10 to the testing matrix.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -40,7 +40,7 @@ jobs:
           python -m pip install dist/*.whl
 
       - name: Run tests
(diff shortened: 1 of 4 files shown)


riddell-stan:
I believe this is failing because of some problem with the very new NumPy wheels for macOS Python 3.10. I suspect this will be resolved within several weeks.

riddell-stan:
It's not NumPy. It's distutils/setuptools/Python. When compiling an extension, Python's default on 3.10 is to target both arm and intel.

We're not trying to cross build right now.

ahartikainen (approved):

riddell-stan:
This still needs work, the clang compiler is not getting used.

PR #597, opened 2022-05-07
feat: Add context to initialization failed error

### LINKED ISSUE #595: Save output of `print`s if init fails
Minimal example:

```
import stan

stan_code = """
parameters { 
  real<upper=0.0> y; 
}
model{
  print(y);
  target += log(y);
}
"""
posterior = stan.build(stan_code, data={})
fit = posterior.sample(num_chains=1, num_samples=1)
```

See forum discussion https://discourse.mc-stan.org/t/debugging-pystan3/27280

When addressing this, it might also be a good idea to see if the verbosity of the error message (on the pystan side) could be reduced.

riddell-stan:
When Stan is unable to initialize a chain, it aborts with an
"Initialization failed" error message. Prior to issuing this error
message, Stan emits potentially useful informational messages, such as
"Rejecting initial value: Log probability evaluates to log(0)".
Previously, these messages were not included in the exception.

Closes #595

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -173,7 +173,7 @@ async def call(
             except json.JSONDecodeError:
                 pass
         # hack to add the info messages to the first argument to the exception, for example
-        # ValueError("Initialization faied.") -> ValueError("Initialization faied. Rejecting initial value: Log probability ...")
+        # ValueError("Initialization failed.") -> ValueError("Initialization faied. Rejecting initial value: Log probability ...")
         if info_messages_for_context and len(exception.args) == 1:
             exception.args = (f"{exception.args[0]} {' '.join(info_messages_for_context)} ...",)
 
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -156,5 +156,26 @@ async def call(
         fh.close()
     httpstan.cache.dump_fit(b"".join(compressed_parts), fit_name)
 
+    # if an exception has already occurred, grab relevant info messages, add as context
+    exception = future.exception()
+    if exception:
+        num_context_messages = 4
+        info_messages_for_context = []
+        import gzip
+        import json
+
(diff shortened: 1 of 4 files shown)


httpstan/services_stub.py
```diff
@@ -156,5 +156,26 @@ async def call(
         fh.close()
     httpstan.cache.dump_fit(b"".join(compressed_parts), fit_name)
 
+    # if an exception has already occurred, grab relevant info messages, add as context
+    exception = future.exception()
+    if exception:
+        num_context_messages = 4
```
ahartikainen, line 162: Where does this come from?

httpstan/services_stub.py
```diff
@@ -156,5 +156,26 @@ async def call(
...
+        jsonlines = gzip.decompress(b"".join(compressed_parts)).decode()
+        for line in jsonlines.split("\n")[:num_context_messages]:
+            try:
+                message = json.loads(line)
+                info_message = message["values"].pop().replace("info:", "")
+                info_messages_for_context.append(info_message.strip())
+            except json.JSONDecodeError:
+                pass
+        # hack to add the info messages to the first argument to the exception, for example
+        # ValueError("Initialization faied.") -> ValueError("Initialization faied. Rejecting initial value: Log probability ...")
```
ahartikainen, line 176:
```suggestion
        # ValueError("Initialization failed.") -> ValueError("Initialization faied. Rejecting initial value: Log probability ...")
```

ahartikainen (approved):

httpstan/services_stub.py
riddell-stan: the future tracks the output of the call to the C++ stan sample function. If that function raises a C++ exception then `exception` here will not be `None`.

riddell-stan:
I'll merge this even though the macOS runners are still working. Seems like GitHub is out of macOS runners at the moment.

PR #600, opened 2022-06-16
[feature request] Add ffi interface for logp-gradient function

aseyboldt:
The current shared object build for each stan model includes a function for computing the logp gradient, but calling this function is relatively slow as it converts the dataset each time. This PR adds a couple of functions that make it possible to call the gradient function from a C or numba library without any python overhead:

```python
model = import_services_extension_module(model_id)
ctx = model.new_logp_ctx(data)
function_address = model.logp_func(ctx)

# later, after we are done
# model.free_logp_ctx(ctx)
```

`function_address` is now a pointer to a C function:
```C
int logp_gradient(size_t ndim, const double *unconstrained_parameters, double *gradient, double *logp, void *ctx)
```
Which is way faster than the previously exported gradient function.

This is useful for implementing different samplers based on stan models (for example the [sampler implemented in rust](https://github.com/aseyboldt/nuts-rs) that I'm working on ;-) ).


riddell-stan:
Feature requests need to be proposed and discussed on the Stan Forums first. The issue tracker is for bugs and confirmed features only.

aseyboldt:
@riddell-stan Sorry, I didn't realize that.
I created a topic here: https://discourse.mc-stan.org/t/feature-proposal-expose-fast-logp-gradient-function-in-httpstan/27856

PR #601, opened 2022-06-16
Give stan libraries higher priority than system libraries

aseyboldt:
When building the stan service module, the linker can pick up system libraries for tbb, because the library include directory is added to the end of the linker command line instead of to the front.
Here I modify the linker command line in distutils to give the stan libraries the highest priority.

An alternative approach would be to add the default distutils library paths to the makefile, so that the stan_service object file uses the system libraries as well.

This *might* be the issue that makes builds fail on conda-forge: https://github.com/conda-forge/httpstan-feedstock/issues/1

Note: I only tested this on linux

### DIFF
--- a/httpstan/build_ext.py
+++ b/httpstan/build_ext.py
@@ -7,15 +7,37 @@ isort:skip_file
 # background: https://bugs.python.org/issue23102
 import setuptools  # noqa: F401
 
-import distutils.command.build_ext
+from distutils.command.build_ext import build_ext
+import distutils.sysconfig
 import distutils.core
 import io
 import os
 import sys
 import tempfile
 from typing import IO, Any, List, TextIO
+from pathlib import Path
 
 from httpstan.config import HTTPSTAN_DEBUG
+
+PACKAGE_DIR = Path(__file__).parent.resolve(strict=True)
+
+# Provide a custom function for building the stan_service module, so
+# that we can control library loading order.
+class stan_build_ext(build_ext):
+    def build_extensions(self) -> None:
+        self.compiler.set_executable(
+            "linker_so",
+            (
+                # The configured linking executable
+                f"{' '.join(distutils.sysconfig.get_config_vars('LDCXXSHARED'))} "
+                # Higher priority for the stan libraries
+                f"{self.compiler.library_dir_option(str(PACKAGE_DIR / 'lib'))} "
+                f"{self.compiler.runtime_library_dir_option(str(PACKAGE_DIR / 'lib'))} "
+                # The remaining default arguments specified in the system config
+                f"{' '.join(distutils.sysconfig.get_config_vars('LDFLAGS'))}"
+            )
+        )
+        return super().build_extensions()
 
 
 def _get_build_extension() -> distutils.command.build_ext.build_ext:  # type: ignore
@@ -24,7 +46,7 @@ def _get_build_extension() -> distutils.command.build_ext.build_ext:  # type: ig
     dist = distutils.core.Distribution()
     # Make sure build respects distutils configuration
     dist.parse_config_files(dist.find_config_files())  # type: ignore
-    build_extension = distutils.command.build_ext.build_ext(dist)  # type: ignore
+    build_extension = stan_build_ext(dist)
     build_extension.finalize_options()
     return build_extension
 
--- a/httpstan/build_ext.py
+++ b/httpstan/build_ext.py
@@ -29,12 +29,12 @@ class stan_build_ext(build_ext):
             "linker_so",
             (
                 # The configured linking executable
-                f"{' '.join(distutils.sysconfig.get_config_vars('LDCXXSHARED'))} "
+                f"{distutils.sysconfig.get_config_vars().get('LDCXXSHARED', '')} "
                 # Higher priority for the stan libraries
                 f"{self.compiler.library_dir_option(str(PACKAGE_DIR / 'lib'))} "
                 f"{self.compiler.runtime_library_dir_option(str(PACKAGE_DIR / 'lib'))} "
                 # The remaining default arguments specified in the system config
-                f"{' '.join(distutils.sysconfig.get_config_vars('LDFLAGS'))}"
+                f"{distutils.sysconfig.get_config_vars().get('LDFLAGS', '')}"
             )
         )
         return super().build_extensions()
--- a/httpstan/models.py
+++ b/httpstan/models.py
@@ -146,16 +146,20 @@ async def build_services_extension_module(program_code: str, extra_compile_args:
     libraries = ["sundials_cvodes", "sundials_idas", "sundials_nvecserial", "tbb"]
     if platform.system() == "Darwin":  # pragma: no cover
         libraries.extend(["tbbmalloc", "tbbmalloc_proxy"])
+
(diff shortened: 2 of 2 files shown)


riddell-stan:
Thanks. This looks interesting. Would be great if it fixed the conda problem.

I might need a bit of time to review this. I need to refresh my memory of how all this stuff works.

riddell-stan:
@aseyboldt Is there a minimal test case you can devise that shows how the existing version is problematic? Could one, say, create a container image with some old debian / ubuntu that has a system TBB that causes problems?

aseyboldt:
I don't use docker much, I hope this does too (not as reproducible I'm afraid...):
To be sure I'd run this in bash, not some other shell, because conda doesn't always set all env vars in other shells (in my experience anyway)
```
conda create -n httpstan-lib-order python poetry tbb sundials tbb tbb-devel
conda activate httpstan-lib-order
git clone https://github.com/stan-dev/httpstan httpstan_tmp
cd httpstan_tmp
make -j4
python -m poetry build
pip install dist/*.whl
rm -rf ~/.cache/httpstan/
python -c "import httpstan.models, asyncio; asyncio.run(httpstan.models.build_services_extension_module('data {}'))"
ldd ~/.cache/httpstan/4.7.2/models/*/*.so
```

This should give an output like this:
```
(httpstan-lib-order) [adr@adr-desktop httpstan_tmp] ldd ~/.cache/httpstan/4.7.2/models/*/*.so
        linux-vdso.so.1 (0x00007ffd5e4c4000)
        libsundials_cvodes.so.6 => /home/adr/miniconda3/envs/httpstan-lib-order/lib/libsundials_cvodes.so.6 (0x00007f84d0ec0000)
        libsundials_idas.so.5 => /home/adr/miniconda3/envs/httpstan-lib-order/lib/libsundials_idas.so.5 (0x00007f84d0e5f000)
        libsundials_nvecserial.so.6 => /home/adr/miniconda3/envs/httpstan-lib-order/lib/libsundials_nvecserial.so.6 (0x00007f84d0e42000)
        libtbb.so.12 => /home/adr/miniconda3/envs/httpstan-lib-order/lib/libtbb.so.12 (0x00007f84d0df2000)
        libstdc++.so.6 => /home/adr/miniconda3/envs/httpstan-lib-order/lib/libstdc++.so.6 (0x00007f84d0c3e000)
        libm.so.6 => /usr/lib/libm.so.6 (0x00007f84d0b19000)
        libgcc_s.so.1 => /home/adr/miniconda3/envs/httpstan-lib-order/lib/libgcc_s.so.1 (0x00007f84d0b00000)
        libc.so.6 => /usr/lib/libc.so.6 (0x00007f84d08f4000)
        /usr/lib64/ld-linux-x86-64.so.2 (0x00007f84d0fcb000)
        libdl.so.2 => /usr/lib/libdl.so.2 (0x00007f84d08ef000)
        libpthread.so.0 => /usr/lib/libpthread.so.0 (0x00007f84d08ea000)
```

Which means we are loading the (incompatible) sundials libraries, and the conda libtbb. But since the `stan_service.o` file is compiled against libtbb from httpstan, this will fail to load due to missing symbols from tbb.

riddell-stan:
Thanks. I can work with this.

aseyboldt:
@riddell-stan I experimented a bit with the conda package, and I think I got it working (and messy for now) on linux and osx, by using the conda dependencies instead of bundled boost, eigen etc...
Windows is still a big mess though.
https://github.com/conda-forge/httpstan-feedstock/pull/14

ahartikainen:
I know there are now unix-sockets on Windows (or what was it we used) but is it already in python? (There was some python issue/pep/their discussion)

aseyboldt:
I created a new [PR](https://github.com/conda-forge/httpstan-feedstock/pull/15) for the conda package.
This changes a bit how httpstan is build in conda:

- Instead of vendoring boost, tbb, eigen and sundials we use the conda packages for those.
- This simplifies the build quite a bit, so that I think we don't need the Makefile anymore. The remaining tasks are done in the build.sh script now.

The reason I'm mentioning this here, is that this also influences how we need to build the model shared libraries later.
If we distribute using pip, we need to link to the libraries in the httpstan tree, but if we distribute with conda, we need to link to the conda libs. So the order of linker flags needs to be different.

Maybe the best way to accomplish this would be to write necessary compiler flags to a file when we build the package (ie in build.py for pip or in build.sh for conda), and load that file in a function like `httpstan._get_compile_flags()`, which we then use in `httpstan.models`?

riddell-stan:
@aseyboldt FYI: in addition to building those shared libraries, the Makefile does build `httpstan/stan_services.o`.

riddell-stan:
@aseyboldt What operating system are you using? I'm trying to reproduce your example in https://github.com/stan-dev/httpstan/pull/601#issuecomment-1159038213 but am failing. I install everything using miniconda but get this when I run ldd:

```
/home/lra/.cache/httpstan/4.7.2/models/msuyi3yt/stan_services_model_msuyi3yt.cpython-310-x86_64-linux-gnu.so:
	linux-vdso.so.1 (0x00007ffcc3fe2000)
	libstdc++.so.6 => /home/lra/miniconda3/envs/httpstan-lib-order2/lib/libstdc++.so.6 (0x00007fd683a3f000)
	libm.so.6 => /lib/x86_64-linux-gnu/libm.so.6 (0x00007fd683945000)
	libgcc_s.so.1 => /home/lra/miniconda3/envs/httpstan-lib-order2/lib/libgcc_s.so.1 (0x00007fd68392b000)
	libc.so.6 => /lib/x86_64-linux-gnu/libc.so.6 (0x00007fd683703000)
	/lib64/ld-linux-x86-64.so.2 (0x00007fd683cfb000)
```

aseyboldt:
@riddell-stan I'm also building the `httpstan/stan_services.o` file in the build.sh in the conda package PR.

Strange, that you got a different result...
Just to make sure, this is without the patch in the PR? The latest release or some other commit?
I'm running this on arch linux.
conda sets some env variables differently depending on the shell. I think I executed it in bash, but I'm not sure, I can retry tomorrow.

riddell-stan:
@aseyboldt Thanks. I don't mind trying this on Arch.

I would like to pin down the problem before looking at potential solutions. For example, does this problem also arise when someone has installed tbb using the package manager (i.e., no anaconda involved)?

riddell-stan:
`readelf` does the trick. I see the order now.

Does setting `runtime_library_dirs` help at all? It would be nice to stick to the traditional distutils calls.

aseyboldt:
I think that only adds arguments to the end, but because the search paths
that conda adds are at the beginning we can not give our paths a higher
priority.

riddell-stan ***@***.***> schrieb am Do., 30. Juni 2022,
12:54:

riddell-stan:
Did some additional digging. Looks like this may be a conda-specific issue.

To make sure, I took an Ubuntu 20.04 LTS host and installed libtbb-dev and libsundials-dev. httpstan ignores the system libraries. You get a clean Library runpath. In this case, I see this:

```
 0x000000000000001d (RUNPATH)            Library runpath: [/home/riddella/tmp/pystan-test2/venv/lib/python3.8/site-packages/httpstan/lib]
```

PR #606, opened 2022-06-28
fix: Sync doc requirements (e.g., Sphinx)

riddell-stan:
sphinx-related requirements are given in two places due to need
to build on readthedocs without installing httpstan.


ahartikainen (approved):

ahartikainen:
I think we need to upgrade our macos runner, I don't think 10.14 is supported anymore.

https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources

riddell-stan:
Thanks for figuring out what was wrong. It wasn't clear to me!

PR #607, opened 2022-06-28
refactor: Improve code readability, init failed fix

riddell-stan:
A recent commit, `feat: Add context to initialization failed error`
included code that was not as clear as it could have been. This
commit improves code legibility.

### DIFF
--- a/httpstan/services_stub.py
+++ b/httpstan/services_stub.py
@@ -158,11 +158,13 @@ async def call(
 
     # if an exception has already occurred, grab relevant info messages, add as context
     exception = future.exception()
-    if exception:
-        num_context_messages = 4
-        info_messages_for_context = []
+    if exception and len(exception.args) == 1:
         import gzip
         import json
+
+        original_exception_message = exception.args[0]  # e.g., from ValueError("Initialization failed.")
+        info_messages_for_context = []
+        num_context_messages = 4
 
         jsonlines = gzip.decompress(b"".join(compressed_parts)).decode()
         for line in jsonlines.split("\n")[:num_context_messages]:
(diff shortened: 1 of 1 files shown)


httpstan/services_stub.py
```diff
@@ -158,12 +158,14 @@ async def call(
 
     # if an exception has already occurred, grab relevant info messages, add as context
     exception = future.exception()
-    if exception:
-        num_context_messages = 4
-        info_messages_for_context = []
+    if exception and len(exception.args) == 1:
```
ahartikainen, line 161: Is there are any cases where the args could be larger than 1?

httpstan/services_stub.py
```diff
@@ -172,10 +174,11 @@ async def call(
                 info_messages_for_context.append(info_message.strip())
             except json.JSONDecodeError:
                 pass
-        # hack to add the info messages to the first argument to the exception, for example
+        # add the info messages to the original exception message. For example,
         # ValueError("Initialization failed.") -> ValueError("Initialization faied. Rejecting initial value: Log probability ...")
```
ahartikainen, line 178:
```suggestion
        # ValueError("Initialization failed.") -> ValueError("Initialization failed. Rejecting initial value: Log probability ...")
```

ahartikainen (approved):
Two comments but otherwise LGTM

httpstan/services_stub.py
riddell-stan: It's certainly possible. Someone could throw `ValueError("foo", "bar")`.

What I really want to do is make sure it's the "Initialization failed" exception, but I don't want to use the string "Initialization failed" because that string is taken from Stan C++ output and could change in the future. For example, they could change it to "Initializing... failed" or something like that.

httpstan/services_stub.py
riddell-stan: Thanks. Fixed. (Will rebase soon.)

httpstan/services_stub.py
ahartikainen: Ok, makes sense

PR #620, opened 2022-09-30
ci: Use newer poetry in CI

riddell-stan:
Poetry 1.2 appeared not too long ago. Use it.

~This needs to be rebased before merging.~


riddell-stan:

PR #622, opened 2022-09-30
docs: Address breaking change in openapi-spec-validator

riddell-stan:
openapi-spec-validator version 0.5 introduced breaking changes.


riddell-stan:
Trivial fix. Merging without review.

PR #626, opened 2022-11-07
ci: Test on Python 3.11

riddell-stan:
Python 3.11 has been released.


riddell-stan:
Looks good.

PR #627, opened 2022-11-08
feat: Update Stan to 2.31.0, bump version

riddell-stan:
Update Stan to 2.31.0.


riddell-stan:
The problem is with setuptools 65.6.x, recently released. Not sure exactly what is going on. I assume other people will encounter it and it will be fixed. Will pin our required setuptools to <65.6, which should not create any problems. We will need to undo the pin as soon as possible.

The relevant error message is `TypeError('dist must be a Distribution instance')`

ahartikainen (approved):

PR #629, opened 2022-11-15
docs: Make clear that wheels need x86-64 arch

riddell-stan:
Make clear that wheels need x86-64 arch and that people with other kinds of CPUs will need to install from source.

Minor edits to align pystan and httpstan docs included here as well.


ahartikainen (approved):
Do we have instructions how one can get the source?

riddell-stan:
That's a good point. Fixed in the newest version.

Issue #631, opened 2022-11-24
Migration from distutils to setuptools incomplete


riddell-stan:
Finish the migration to setuptools and use the public API. Currently httpstan is still using distutils directly, e.g., `import setuptools._distutils.command.build_ext`

For migration advice see <https://setuptools.pypa.io/en/latest/deprecated/distutils-legacy.html>

PR #633, opened 2022-12-05
Finish migration to setuptools

riddell-stan:
Finish migration from distutils to setuptools. Distutils is being deprecated by Python.

There are minor differences between setuptools and distutils. These differences are not well-documented.

This change appears to make things work.

A better approach would be to subclass build_ext, following the
guidelines in the setuptools documentation:

https://setuptools.pypa.io/en/latest/userguide/extension.html

Closes https://github.com/stan-dev/httpstan/issues/631

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -37,7 +37,7 @@ jobs:
       # export test dependencies from pyproject.toml, install them
       - name: Install dependencies
         run: |
-          poetry export -f requirements.txt --without-hashes --dev -o requirements.txt \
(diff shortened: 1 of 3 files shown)


riddell-stan:
For anyone interested, here's a complex example of someone customizing build_ext: https://github.com/ratt-ru/montblanc/blob/aaf23809c890de13e76e18bf69a174236b4eb5bb/install/tensorflow_ops_ext.py

ahartikainen (approved):

PR #642, opened 2023-09-08
feat: Update Stan to 2.33.0, bump version

riddell-stan:
Update Stan to 2.33.0, bump httpstan version.

PR #644, opened 2023-09-09
docs: pin jsonschema

ahartikainen:
Pin jsonschema to 4.17.3 to be compatible with sphinx.


ahartikainen:
We probably want to leave a comment there why are pinning jsonschema

    # compatibility with sphinx 4.2

But once we update sphinx we can remove the whole line. Should add a comment about it too?

riddell-stan:
Have a look at https://github.com/stan-dev/httpstan/pull/642 -- I think I solved the issue with building docs. The compatibility issue involves, I think, sphinxcontrib-openapi and jsonschema. I don't think sphinx itself is implicated.

ahartikainen:
Good catch

ahartikainen:
Fix in #642

PR #646, opened 2023-10-06
Add a "clean" rule for Makefile

tncowart:
A "clean" rule will make fully resetting the state of a repository post- build easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,6 +9,8 @@
 # this Makefile copy libraries built by the other Makefile into their
 # httpstan-specific directories.
 
+.PHONY: clean default
+
 PYBIND11_VERSION := 2.9.2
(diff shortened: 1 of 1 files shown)


riddell-stan:
Since all development happens in git, I think `git clean` makes this rule unnecessary.

tncowart:
What about people who download the tarball to build from source?

tncowart:
In particular, from the installation documents: `Download the source code associated with a release from https://github.com/stan-dev/httpstan/tags. Extract the contents of the zip or tar.gz release bundle.` So not everyone using the source code will have the code set up in a git repository.

Also, people might be developing and may not want to delete all untracked files. If, for instance they have new files they've created.

PR #648, opened 2023-10-16
Minimal m1 build

pdjames:
Hi team,

I've created a minimal changes branch to enable M1 builds using GitHub's recently added M1 runners.
- steps in the build task are the same
- M1 runners currently have a price attached, but it's not huge and these builds are in the order ~$1, so not terrible but developing changes against them might not be fun, try on a private machine first. 
- I have frozen about 20 versions of dev dependencies in this commit.
- hopefully dev dependencies aren't seen as too critical to have bleeding edge versions.

There were a few discoveries a long the way, the GitHub workflow changes here are minimal, there was a lot of time spent discovering a combination of dev dependencies which pass muster on a variety of os & arch combinations. The package solver didn't always get it right wrt what would run, just what would install. 

I think this link is public, but let me know if you can't see it;
https://github.com/pdjames/httpstan_build/actions/runs/6530495372
This shows the exact same commit just without the dev dependency changes is failing, in this case it's autoapi that fails, though I have walked up that tree and arrived at what is committed in this PR.

Let me know if you have any questions, 
Thanks!

Peter


riddell-stan:
So no changes to the Makefile or build steps, right?

I've created a simpler PR https://github.com/stan-dev/httpstan/pull/649 . It doesn't appear to be working at the moment. I'll try to sort out the billing next week.

pdjames:
I recognise those apispec errors, in this pull request I've worked around the sphinx versions by stepping versions backwards.

I have another PR here https://github.com/stan-dev/httpstan/pull/650 which has the versions stepping forwards and all the existing and the new builds pass.

I'll close this PR as the newer is simpler.

Correct, no changes to the makefile needed.

PR #653, opened 2023-12-14
chore: Update supported Pythons, drop 3.8

riddell-stan:


riddell-stan:
Something breaks with Python 3.12, alas. Holding off on that for now.

PR #663, opened 2025-02-05
docs: Note build requirement for poetry<2

riddell-stan:
Note build requirement for poetry<2


riddell-stan:
This is a docs-only change. I'll merge it despite the mypy failures.

PR #666, opened 2026-07-02
Update Stan to version 2.36.0

riddell-stan:
With one exception, routine updates/fixes. The exception is the addition of a C++ unix socket writer to replace boost/asio's. Stan Math 2.36.0+ no longer vendors boost/asio so we can't use it (easily) for this purpose.


riddell-stan:
Seeing no objection, I'll merge this.

PR #667, opened 2026-07-29
Update Stan to 2.39.0

riddell-stan:
Update Stan to 2.39.0 (and add commits for earlier versions)


bob-carpenter:
Nice!  Thanks for updating, @riddell-stan.

